PR #4, opened 2019-12-16
LVFM-328: bssrf section address and size read from elf file

hcleonis:
Additional ram page address & size now read from ELF file. If provided on the CLI by the user, parameters -q & -r will supersede the values from the ELF file.

### DIFF
--- a/speculos.py
+++ b/speculos.py
@@ -49,8 +49,10 @@ def get_elf_infos(app_path):
             print('[-] failed to find _estack/END_STACK symbol', file=sys.stderr)
             sys.exit(1)
         estack = sym_estack[0]['st_value']
+        supp_ram = elf.get_section_by_name('.rfbss')
+        ram_addr, ram_size = (supp_ram['sh_addr'], supp_ram.data_size) if supp_ram is not None else (0, 0)
     stack_size = estack - stack
-    return hex(sh_offset), hex(sh_size), hex(stack), hex(stack_size)
+    return hex(sh_offset), hex(sh_size), hex(stack), hex(stack_size), hex(ram_addr), hex(ram_size)
 
 def run_qemu(s1, s2, app_path, libraries=[], seed=DEFAULT_SEED, debug=False, trace_syscalls=False, model=None, rampage=None, pagesize=None, sdk_version="1.5"):
     args = [ 'qemu-arm-static' ]
@@ -63,9 +65,6 @@ def run_qemu(s1, s2, app_path, libraries=[], seed=DEFAULT_SEED, debug=False, tra
     if trace_syscalls:
         args += [ '-t' ]
 
-    if rampage is not None and pagesize is not None:
-        args += [ '-r' , rampage ] + [ '-s' , pagesize ]
-
     if model is not None:
         args += ['-m', model]
 
@@ -73,8 +72,15 @@ def run_qemu(s1, s2, app_path, libraries=[], seed=DEFAULT_SEED, debug=False, tra
 
     for lib in [ f'main:{app_path}' ] + libraries:
         name, lib_path = lib.split(':')
-        load_offset, load_size, stack, stack_size = get_elf_infos(lib_path)
+        load_offset, load_size, stack, stack_size, ram_addr, ram_size = get_elf_infos(lib_path)
         args.append(f'{name}:{lib_path}:{load_offset}:{load_size}:{stack}:{stack_size}')
+
+    if rampage is not None and pagesize is not None:
+        print('[+] using user-provided additional RAM page')
+        args += [ '-r' , rampage ] + [ '-s' , pagesize ]
+    else:
+        print('[*] using probed additional RAM page')
+        args += [ '-r' , ram_addr ] + [ '-s' , ram_size]
 
     pid = os.fork()
     if pid != 0:
@@ -116,7 +122,7 @@ if __name__ == '__main__':
     parser.add_argument('-x', '--text', action='store_true', help="Text UI (implies --headless)")
     parser.add_argument('-y', '--keymap', action='store', help="Text UI keymap in the form of a string (e.g. 'was' => 'w' for left button, 'a' right, 's' both). Default: arrow keys")
     parser.add_argument('-r', '--rampage', required='--pagesize' in sys.argv or '-q' in sys.argv, type=lambda x: f"{int(x,16):X}", action='store', help='Hex address (prefixed or not with \'0x\') of one additional RAM page available to the app. Requires -q.')
-    parser.add_argument('-q', '--pagesize', required='--rampage' in sys.argv or '-r' in sys.argv, type=lambda x: f"{int(x,16):X}", action='store', help='Byte size (in decimal or in hex if prefixed with \'0x\') of the additional RAM page available to the app. Requires -r.')
+    parser.add_argument('-q', '--pagesize', required='--rampage' in sys.argv or '-r' in sys.argv, type=lambda x: f"{int(x,16):X}", action='store', help='Byte size (in hexadecimal, prefixed or not with \'0x\') of the additional RAM page available to the app. Requires -r.')
     group_progressive = parser.add_mutually_exclusive_group()
     group_progressive.add_argument('-p', '--progressive', action='store_true', help='Enable step-by-step rendering of graphical elements (default for Blue, can be disabled with -P)')
     group_progressive.add_argument('-P', '--flushed', action='store_true', help='Disable step-by-step rendering of graphical elements (default for Nano S)')
--- a/speculos.py
+++ b/speculos.py
@@ -52,7 +52,7 @@ def get_elf_infos(app_path):
         supp_ram = elf.get_section_by_name('.rfbss')
         ram_addr, ram_size = (supp_ram['sh_addr'], supp_ram.data_size) if supp_ram is not None else (0, 0)
     stack_size = estack - stack
-    return hex(sh_offset), hex(sh_size), hex(stack), hex(stack_size), hex(ram_addr), hex(ram_size)
+    return sh_offset, sh_size, stack, stack_size, ram_addr, ram_size
 
 def run_qemu(s1, s2, app_path, libraries=[], seed=DEFAULT_SEED, debug=False, trace_syscalls=False, model=None, rampage=None, pagesize=None, sdk_version="1.5"):
     args = [ 'qemu-arm-static' ]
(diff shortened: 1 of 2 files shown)


speculos.py
```diff
@@ -63,19 +65,23 @@ def run_qemu(s1, s2, app_path, libraries=[], seed=DEFAULT_SEED, debug=False, tra
...
     for lib in [ f'main:{app_path}' ] + libraries:
         name, lib_path = lib.split(':')
-        load_offset, load_size, stack, stack_size = get_elf_infos(lib_path)
+        load_offset, load_size, stack, stack_size, ram_addr, ram_size = get_elf_infos(lib_path)
         args.append(f'{name}:{lib_path}:{load_offset}:{load_size}:{stack}:{stack_size}')
 
+    if rampage is not None and pagesize is not None:
+        print('[+] using user-provided additional RAM page')
+        args += [ '-r' , rampage ] + [ '-s' , pagesize ]
+    else:
```
Gbillou: if there's no bssrf, i.e. for nanos application we should not add these parameters

hcleonis: OK

hcleonis: Fixed. If no such page was declared when the app was linked (i.e. any Nano or Blue app built against an SDK the linker script of which did not declare such page e.g. SDK 1.5), then auto-probing returns 0 and no additional page is used.

greenknot:
The logic seems flawed to me: `ram_addr` and `ram_size` are retrieved for the main ELF, but also for each library. Eventually, the value of the last specified library are used to build the `-r` and `-s` arguments.

hcleonis:
> The logic seems flawed to me: `ram_addr` and `ram_size` are retrieved for the main ELF, but also for each library. Eventually, the value of the last specified library are used to build the `-r` and `-s` arguments.

You're right. Bummer! I'll fix it.

speculos.py
```diff
@@ -128,29 +160,40 @@ def run_qemu(s1, s2, app_path, libraries=[], seed=DEFAULT_SEED, debug=False, tra
     args = parser.parse_args()
     args.model.lower()
 
-    if not (args.flushed or args.progressive):
-        if args.model == 'blue':
-            args.progressive = True
+    try:
```
greenknot: I wouldn't put everything under try/except because part of this code don't raise exceptions. Also I would directly exit with an error message instead of raising `ValueError`

speculos.py
```diff
@@ -128,29 +160,40 @@ def run_qemu(s1, s2, app_path, libraries=[], seed=DEFAULT_SEED, debug=False, tra
...
+                args.flushed = True
+
+        if args.model != 'blue' and (args.rampage != None or args.pagesize != None):
+            # Extra RAM page only allowed on Blue device
+            raise ValueError(f"Extra RAM page arguments -r (--rampage) & -q (--pagesize)) require '-m blue'")
+
+        if args.text:
+            if args.model != 'nanos':
+                raise ValueError(f"Unsupported model '{args.model}' with -x")
+            if args.ontop:
```
greenknot: `parser.add_mutually_exclusive_group()` maybe?

speculos.py
```diff
@@ -63,18 +97,20 @@ def run_qemu(s1, s2, app_path, libraries=[], seed=DEFAULT_SEED, debug=False, tra
...
     for lib in [ f'main:{app_path}' ] + libraries:
         name, lib_path = lib.split(':')
-        load_offset, load_size, stack, stack_size = get_elf_infos(lib_path)
-        args.append(f'{name}:{lib_path}:{load_offset}:{load_size}:{stack}:{stack_size}')
+        load_offset, load_size, stack, stack_size, ram_addr, ram_size = get_elf_infos(lib_path)
+        # Since binaries loaded as libs could also declare extra RAM page(s), collect them all
+        if (ram_addr, ram_size) != (0, 0):
+            extra_ram.append({'addr':ram_addr, 'size':ram_size})
+        args.append(f'{name}:{lib_path}:{hex(load_offset)}:{hex(load_size)}:{hex(stack)}:{hex(stack_size)}')
+
```
greenknot: I guess it's clearer to add the condition `if model == 'blue'` here.

speculos.py
```diff
@@ -49,11 +49,45 @@ def get_elf_infos(app_path):
...
+            origin=['+', 'user-provided']
+
+        if extra_ram is not None:
+            if  len(extra_ram) > 1:     # then all pages shall be identical
+                # Remove identical extra ram params,the python way. Throw an error if resulting list length is not 1
+                extra_ram = [dict(s) for s in set(tuple(d.items()) for d in extra_ram)]
+                if len(extra_ram) > 1:
+                    raise ValueError("Different extra RAM pages for main app and/or libraries!")
+            page = extra_ram[0]
+            print(f'[{origin[0]}] using {origin[1]} {page["size"]:d}-byte additional RAM page @0x{page["addr"]:08x}')
```
greenknot: No need to use string formatting because `hex` returns a string

speculos.py
```diff
@@ -63,18 +97,20 @@ def run_qemu(s1, s2, app_path, libraries=[], seed=DEFAULT_SEED, debug=False, tra
...
 
     args += [ '-k', str(sdk_version) ]
 
     for lib in [ f'main:{app_path}' ] + libraries:
         name, lib_path = lib.split(':')
-        load_offset, load_size, stack, stack_size = get_elf_infos(lib_path)
-        args.append(f'{name}:{lib_path}:{load_offset}:{load_size}:{stack}:{stack_size}')
+        load_offset, load_size, stack, stack_size, ram_addr, ram_size = get_elf_infos(lib_path)
+        # Since binaries loaded as libs could also declare extra RAM page(s), collect them all
+        if (ram_addr, ram_size) != (0, 0):
```
greenknot: If `extra_ram` is a list of tuples, the rest of the code will be much simpler.

speculos.py
```diff
@@ -115,8 +147,8 @@ def run_qemu(s1, s2, app_path, libraries=[], seed=DEFAULT_SEED, debug=False, tra
     group.add_argument('-o', '--ontop', action='store_true', help='The window stays on top of all other windows')
     parser.add_argument('-x', '--text', action='store_true', help="Text UI (implies --headless)")
     parser.add_argument('-y', '--keymap', action='store', help="Text UI keymap in the form of a string (e.g. 'was' => 'w' for left button, 'a' right, 's' both). Default: arrow keys")
-    parser.add_argument('-r', '--rampage', required='--pagesize' in sys.argv or '-q' in sys.argv, type=lambda x: f"{int(x,16):X}", action='store', help='Hex address (prefixed or not with \'0x\') of one additional RAM page available to the app. Requires -q.')
```
greenknot: Please use `type=lambda x: int(x,16)` to avoid a conversion to int in `sanitize_ram_args()`

speculos.py
hcleonis: You don't like the exception mechanism , do you? ;-)
- "try" clause moved closer to where exceptions could be raised (e.g. run_qemu() which uses functions from "os" module which in turn can raise a FileNotFound exception for whatever reason). 
- Replaced the "raise ValueError" statements w/ print()/exit() code.

speculos.py
hcleonis: Won't change: 
Actually argument `--ontop` is already in a mutually exclusive group with `--headless`. And based on its description, the `--text` argument overrides `--headless`. So it cannot be added to the same group. And I don't think adding a 2nd mutualy exclusive group with `--ontop` and `--text` would work since `--ontop ` is already in another group.

speculos.py
hcleonis: Added `model == blue` to the if clause, but did not remove the rest of the clause since it doesn't make sense to consider a zero-byte page starting at address 0.

speculos.py
hcleonis: Removed `:d` formatting for page size. Replaced `:08x` formatting with call to hex() for page address.

hcleonis: It would actually simplify only the following line in `sanitize_ram_args()` function:
```                extra_ram = [dict(s) for s in set(tuple(d.items()) for d in extra_ram)]```
would be changed to:
```                extra_ram = [tuple(s) for s in set(d for d in extra_ram)]```
or something close to that. But in the rest of the code we'd have to address data in extra_ram through indexes. I don't think this change is worth the loss of readability.

File: speculos.py (hunk 6 above)
Comment (hcleonis) [reply speculos.py]: unless I'm missing something here, it would actually simplify only the following line in `sanitize_ram_args()` function:
```                extra_ram = [dict(s) for s in set(tuple(d.items()) for d in extra_ram)]```
would be changed to:
```                extra_ram = [tuple(s) for s in set(d for d in extra_ram)]```
or something close to that. But in the rest of the code we'd have to address data in extra_ram through indexes. I don't think this change is worth the loss of readability.

File: speculos.py (hunk 7 above)
Comment (hcleonis) [reply speculos.py]: Done.

File: speculos.py (hunk 4 above)
Comment (greenknot) [reply speculos.py]: Sorry, I probably commented the wrong line. I would move the `if model == 'blue':` clause outside of `sanitize_ram_args()`:

```
if model == 'blue':
    args.extend(sanitize_ram_args(model, rampage, pagesize, extra_ram))
```

File: speculos.py (hunk 2 above)
Comment (greenknot) [reply speculos.py]: Actually I don't see the point of catching an exception if no action is made but writing the error message and exiting :)

File: speculos.py (hunk 5 above)
Comment (greenknot) [reply speculos.py]: Same for the next line please :) (`f'{hex(page["addr"])}'` and `f'{hex(page["size"])}'`)

File: speculos.py (hunk 4 above)
Comment (hcleonis) [reply speculos.py]: done.

File: speculos.py (hunk 2 above)
Comment (hcleonis) [reply speculos.py]: This was to make the error printing cleaner rather than having the full exception  stack dumped on the user in case of network or file error. But we can live w/ exception stack, right?
=> removed the try/except clause entirely.

File: speculos.py (hunk 5 above)
Comment (hcleonis) [reply speculos.py]: done

File: speculos.py (hunk 4 above)
Comment (greenknot) [reply speculos.py]: The condition `if model == 'blue':` in `sanitize_ram_args()` isn't required anymore and can be removed. Thanks!

File: speculos.py [hunk 8]
```diff
@@ -92,13 +129,9 @@ def run_qemu(s1, s2, app_path, libraries=[], seed=DEFAULT_SEED, debug=False, tra
     os.environ['SPECULOS_SEED'] = binascii.hexlify(seed).decode('ascii')
 
     #print('[*] seproxyhal: executing qemu', file=sys.stderr)
-    try:
-        os.execvp(args[0], args)
-    except FileNotFoundError:
-        print('[-] failed to execute qemu: "%s" not found' % args[0], file=sys.stderr)
-        sys.exit(1)
```
greenknot: This shouldn't be removed.

speculos.py
hcleonis: done

speculos.py
hcleonis: restored

PR #8, opened 2020-01-02
Connect/Disconnect on each request

jleni:
The purpose is to avoid restarting the proxy each time a qemu session is closed or restarted.

(also fixes a small problem with subrepos that were not committed correctly)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,0 @@
-[submodule "openssl"]
-	path = openssl
-	url = https://github.com/openssl/openssl.git
-[submodule "tests/cmocka"]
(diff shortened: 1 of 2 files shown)


greenknot:
Great, thanks!

Actually, the submodules aren't use anymore since a919bfe877d2e68f321cb20ed75bdddba6f9f049; but we forgot to remove the `.gitmodules` file. Unless I misunderstood something, could you remove this file and also revert 5d10f7609c7a4b3493df0a9fc8908d4e5677bccc?

greenknot:
Squashed and merged manually, thanks again!

PR #13, opened 2020-01-10
Add BIP39 hex seed support

TamtamHero:
`--seed` arg actually expected a BIP39 mnemonic, not a seed.
This PR allows using real BIP39 seeds with this same arg, by prefixing it with `hex:`

eg:
`--seed abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon about`
is equivalent to 
`--seed hex:5eb00bbddcf069084889a8ab9155568165f5c453ccb85e70811aaed6f6da5fc19a5ac40b389cd370d086206dec8aa6c43daea6690f20ad3d8d48b2d2ce9e38e4`

Use case:
As going from mnemonic to seed is done through a one way function (PBKDF2), there are a few cases where we only have access to the seed, but we can't recompute the source mnemonic from it.

### DIFF
--- a/speculos.py
+++ b/speculos.py
@@ -89,7 +89,7 @@ def run_qemu(s1, s2, app_path, libraries=[], seed=DEFAULT_SEED, debug=False, tra
     os.dup2(s1.fileno(), sys.stdin.fileno())
 
     # handle both BIP39 mnemonics and hex seeds
-    if "hex:" in seed:
+    if seed.startswith("hex:"):
         seed = bytes.fromhex(seed[4:])
     else:
         seed = mnemonic.Mnemonic.to_seed(seed)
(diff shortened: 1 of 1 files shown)


speculos.py
```diff
@@ -88,7 +88,12 @@ def run_qemu(s1, s2, app_path, libraries=[], seed=DEFAULT_SEED, debug=False, tra
     # replace stdin with the socket
     os.dup2(s1.fileno(), sys.stdin.fileno())
 
-    seed = mnemonic.Mnemonic.to_seed(seed)
+    # handle both BIP39 mnemonics and hex seeds
+    if "hex:" in seed:
```
greenknot: I think something like `if seed.startswith("hex":)` might be better.

TamtamHero: Indeed !
Fixed.

greenknot:
Thanks!

PR #15, opened 2020-01-16
seproxyhal: implement sys_io_seph_is_status_sent

### LINKED ISSUE #11: Concatenation of APDU messages is not supported
Longer APDU messages are typically read into `G_io_apdu_buffer` by concatenating multiple smaller messages that are received in `G_io_seproxyhal_spi_buffer`. It seems like this functionality relies on `io_seproxyhal_spi_is_status_sent` to return a non-zero value when more data is available (i.e. sent on SPI but not yet read into `G_io_seproxyhal_spi_buffer`). I think that [this code](https://github.com/LedgerHQ/nanos-secure-sdk/blob/1f2706941b68d897622f75407a868b60eb2be8d7/src/os_io_seproxyhal.c#L1186-L1194) may be related.

The current implementation of this function always returns zero, which effectively disables support for message concatenation (see [/src/emu_seproxyhal.c](https://github.com/LedgerHQ/speculos/blob/be45a25e5af7c52b733d9deb781ca3bc656cf600/src/emu_seproxyhal.c#L57)).

With message concatenation disabled, received messages risk to be cut off if `G_io_apdu_buffer` is larger than `G_io_seproxyhal_spi_buffer`. In these cases, the cut off part is filled with zeroes.

The problem can be mitigated on the app side by defining `IO_SEPROXYHAL_BUFFER_SIZE_B` to be sufficiently large. However, the current implementation creates confusion as to why apps behave the way they do when tested using Speculos.

greenknot:
Fix #11.

### DIFF
--- a/src/emu_seproxyhal.c
+++ b/src/emu_seproxyhal.c
@@ -4,6 +4,13 @@
 #include <unistd.h>
 
 #include "emulate.h"
+#include "exception.h"
+
+#define SEPROXYHAL_TAG_STATUS_MASK  0x60
+
+static bool status_sent = false;
+static uint8_t last_tag;
+static size_t next_length;
 
 static ssize_t read_helper(int fd, void *buf, size_t count)
 {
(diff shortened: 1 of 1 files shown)


greenknot:
FTR, it doesn't fix #11.

PR #16, opened 2020-01-17
Docker

guycaprio:
Container is running, websockify is responding but I did not have the tools & knowledge to test fully vnc & apdu


.dockerignore
```diff
@@ -1,3 +1,2 @@
-apps
```
greenknot: I would prefer having this folder (`app`) ignored to prevent developers for releasing private apps.

Apps can still be used from containers thanks to [docker --volumes](https://docs.docker.com/storage/volumes/). I guess this requires additional minor changes in `doc/usage.md` and `run.sh`.

greenknot:
Thanks a lot!

.dockerignore
guycaprio: ✔restored .dockerignore
✔doc/usage.md
✔docker-compose.yml
✘ run.sh (no changes required, the whole apps/ directory is mounted)

PR #23, opened 2020-02-10
Removed websockify from docker

guycaprio:


meriadec:
this was used by WeBlue project https://github.com/LedgerHQ/ledger-vault-front/pull/1154
removing this breaks vnc support ¯\\\_(ツ)_/¯ SFOL

gre:
@meriadec I think this was a good direction tho. this project should expose the minimal for all teams to depends on and uses with different usecases. it's less things for speculos team to maintain and more unix fridently with the recent iteration. (you can literally inline everything in params instead of defining a dockerfile)

PR #26, opened 2020-02-14
Add new entrypoint for weblue using websocket redirection

YBadiss:
WeBlue requires a different docker entrypoint to run speculos, expecting a few env variables to customize the run.
This new entrypoint `scripts/weblue_entrypoint.sh` is kept separate from the "normal" run of speculos, so we are not disturbing existing users with our special flow.

In addition, it handles APDUs and VNC over websockets, so I am adding back websockify to the python dependencies.

### DIFF
--- a/scripts/weblue_entrypoint.sh
+++ b/scripts/weblue_entrypoint.sh
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+set -euxo pipefail
+
+APDU_PORT=40000
+VNC_PORT=41000
+
+websockify 50000 localhost:${APDU_PORT} --cert ./certificate/ws.pem &  # for apdu
+websockify 60000 localhost:${VNC_PORT} --cert ./certificate/ws.pem &  # for vnc
+exec ./speculos.py ${DEBUG_MODE:-} -m ${DEVICE_MODEL} --sdk ${SDK_VERSION} --seed "${DEVICE_SEED}" --rampage ${RAMPAGE} --display headless --vnc-port ${VNC_PORT} --apdu-port ${APDU_PORT} ./apps/${APP_FILE}
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 *.o
 .cache/
 .idea/
+.DS_Store
 __pycache__/
 apps/*vault*
 build/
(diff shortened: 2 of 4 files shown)


scripts/weblue_entrypoint.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+set -euxo pipefail
+
+APDU_PORT=40000
+VNC_PORT=41000
+
+websockify 50000 localhost:${APDU_PORT} --cert ./certificate/ws.pem &  # for apdu
+websockify 60000 localhost:${VNC_PORT} --cert ./certificate/ws.pem &  # for vnc
+exec ./speculos.py ${DEBUG_MODE:-} -m ${DEVICE_MODEL} --sdk ${SDK_VERSION} --seed "${DEVICE_SEED}" --rampage ${RAMPAGE} --display headless --vnc-port ${VNC_PORT} --apdu-port ${APDU_PORT} ./apps/${APP_FILE}
```
gre: my 2 cents:

this should not live in this repository. but maybe i'm wrong. basically with recent @guycaprio iteration we figured out a way to expose speculos params directly through a `docker run` . it sounds like minimal enough to not have to do too much specifics boilerplate in place. (websockify was actually removed recently in the purpose of the simplification)

otherwise, the specific should live in each projects (like internal tests, weblue,...) to lighter the maintenance of the simulator itself.

guycaprio: I agree with that, maybe the good way for _weblue_ is to create another container (or a simple script?) and set up all services in `docker-compose.yml` to make it easy to instantiate.

Source: https://docs.docker.com/config/containers/multi-service_container/
> It is generally recommended that you separate areas of concern by using one service per container.

Anyway, the PR is not breaking anything, I approve and let the maintainer decides what to do (cc @greenknot).

YBadiss: Yes, you are both right, weblue will build its own docker image based on the speculos build. Just tried it out and with the extra websockify dep and overridden entrypoint this works fine 👌 

It makes way more sense, thanks for the comments!

YBadiss:
Following the comments from @gre and @guycaprio , keeping this internal to weblue.

greenknot:
:+1:

PR #35, opened 2020-03-12
Allow CORS in ledger-live-http-proxy

vgrichina:
This makes it possible to access this server from web page running on different origin (host / port).


greenknot:
Thanks!

PR #48, opened 2020-04-01
Add Docker Hub auto-publish to CI

TamtamHero:


.github/workflows/continuous-integration-workflow.yml
```diff
@@ -33,3 +33,13 @@ jobs:
       run: |
         make -C build/ test
         pytest
+
+    - name: Publish to DockerHub
+      uses: docker/build-push-action@v1
+      with:
+        repository: ledgerhq/speculos
+        dockerfile: CI.Dockerfile
```
YBadiss, line 41: Can't we have a single Dockerfile? They seem to be slightly different in there builds, that could lead to weird incompatibilities in the future.

greenknot: Indeed. There are 3 Dockerfiles:

- `build.Dockerfile`: used by the CI to build speculos. This image is on [docker hub](https://hub.docker.com/repository/docker/ledgerhq/speculos-builder) 
- `CI.Dockerfile`: runs speculos (the image embeds the build artifacts from the CI build); should be available on docker hub too
- `Dockerfile`: allows to build and run speculos

IIUC, `Dockerfile` can be replaced by `CI.Dockerfile`? Did I miss something? What do you think about it @nabil-brn ?

nabil-brn: I agree with @YBadiss, what about using environments and conditions in Dockerfile and `--build-arg` at run time ?

What is the best for you ?

greenknot: Ack, thanks!

PR #56, opened 2020-04-28
add support for HDW_NORMAL/Ed25519 in os_perso_derive_node_with_seed_key

greenknot:


src/emu_os_bip32.c
```diff
@@ -147,6 +168,9 @@ unsigned long sys_os_perso_derive_node_bip32(cx_curve_t curve, const uint32_t *p
       memmove(tmp + 1, tmpPrivateChain, 32);
     }
     else {
+      if (curve == CX_CURVE_Ed25519) {
+        errx(1, "os_perso_derive_node_bip32: invalid path");
```
jleni, line 172: We are hitting this error here. However, AFAIK, this restriction applied to HDW_ED25519_SLIP10 and not HDW_NORMAL.

tests/syscalls/test_bip32.c
```diff
@@ -11,59 +11,195 @@
 #include "emu_os_bip32.h"
 #include "emulate.h"
 
+#define ARRAY_SIZE(x)  (sizeof(x) / sizeof(x[0]))
+#define MAX_CHAIN_LEN  5
```
jleni, line 15: While this is ok for BIP44, we have some projects that may require longer paths. Would it make sense to extend this to 10?

greenknot: That's just the max length used by these tests, not the syscall implementation limit ;)

src/emu_os_bip32.c
greenknot: Ack, thanks.

jleni:
@greenknot I saw a fair amount of work on ed25519 in other commits..
It this PR obsolete and support for ed25591 has improved in the main branch?

src/emu_os_bip32.c
```diff
@@ -121,128 +205,266 @@ static size_t get_seed_from_env(const char *name, uint8_t *seed, size_t max_size
...
+    cx_math_add(ZL, ZL, ZL, 32);
+    cx_math_add(ZL, ZL, kLP, 32);
+    be2le(ZL, 32);
+    be2le(kLP, 32);
+    // kR = Zr + kRP
+    le2be(ZR, 32);
+    le2be(kRP, 32);
+
+    // cx_math_add(ZR, ZR, kRP, 32) can return invalid result when an overflow
+    // occurs
```
niooss-ledger, line 277: The comment would be clearer if something like this is added: " so perform addition on 33 bytes and truncate the result to 32 bytes"

greenknot: Fixed. Thanks!

PR #59, opened 2020-05-02
Missing left/right glyphs

jleni:
We use these two glyphys a lot. It would be nice to have them included.


greenknot:
Cool, thanks!

PR #62, opened 2020-05-06
map down key as a shortcut for pressing both left and right keys

### LINKED ISSUE #61: Button Handling OSX
It seems in OSX speculos doesn't handle the `BUTTON_EVT_RELEASED | BUTTON_LEFT | BUTTON_RIGHT` event when I press down both the `left` and `right` keyboard keys. 

Is there anyway to fix this or to emulate this event programatically? Thanks.

greenknot:
Fix #61.

### DIFF
--- a/doc/usage.md
+++ b/doc/usage.md
@@ -26,7 +26,8 @@ For more options, pass the `-h` or `--help` flag.
 
 #### Keyboard control
 
-- The keyboard left and right arrow keys are use instead of the Nano buttons.
+- The keyboard left and right arrow keys are used instead of the Nano buttons.
+  The down arrow can also be used as a more convenient shortcut.
 - The `Q` key exits the application.
 
 #### Display
--- a/mcu/screen.py
+++ b/mcu/screen.py
@@ -83,6 +83,9 @@ class Screen(Display):
             buttons = { Qt.Key_Left: BUTTON_LEFT, Qt.Key_Right: BUTTON_RIGHT }
             # forward this event to seph
             self.seph.handle_button(buttons[key], pressed)
+        elif key == Qt.Key_Down:
+            self.seph.handle_button(BUTTON_LEFT, pressed)
+            self.seph.handle_button(BUTTON_RIGHT, pressed)
         elif key == Qt.Key_Q and not pressed:
             self.app.close()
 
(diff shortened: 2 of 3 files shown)


niooss-ledger:
If this feature solves the issue, could it be documented in https://github.com/LedgerHQ/speculos/blob/master/doc/usage.md#keyboard-control ? (I find it useful, and it would be helpful to have it in the usage doc)

mul1sh:
@greenknot thanks for the quick feedback, but unfortunately it still does not work when I press the down button and try to handle the `BUTTON_EVT_RELEASED | BUTTON_LEFT | BUTTON_RIGHT` event.

Though when I pressed the `left` + `right` buttons in the main menu of the app i.e. in About section, it opens the menu as expected. It only fails to work with the above event only.

greenknot:
@niooss-ledger done

@mul1sh Indeed, the down key wasn't handled by the VNC server. The last commit should fix that, can you give it a try? Please note that the VNC server needs to be built again (`make -C build/vnc`).

mul1sh:
@greenknot thanks again for the fast response, I truly appreciate :).

Unfortunately the above event is still not handled in speculos when I press the down key. It seems to fire both the `left` and `right` buttons at the same time so the text being scrolled is moved a bit  below to the left or right side as shown below.
down_key_event

When I press the down command in the main menu of the app though, it works perfectly as expected as shown below.
main_menu

So not really sure why its not working with the `BUTTON_EVT_RELEASED | BUTTON_LEFT | BUTTON_RIGHT` event as expected. Same thing happens when I press both the left and right buttons btw.

greenknot:
IIUC, pressing both left and right keys (or using the down key) works in the main menu of the app, but doesn't work elsewhere?

mul1sh:
Yes correct

greenknot:
It looks like a bug in you code to me :) Can you share it?

mul1sh:
@greenknot i'm actually running the default sia [boilerplate app](https://github.com/LedgerHQ/app-sia):) 

It fails at this line https://github.com/LedgerHQ/app-sia/blob/master/src/signHash.c#L210 in OSX. For some reason this event is never "called" when I fire either the `down` button or the `left`/`right` buttons.

greenknot:
While it doesn't fix #61, this PR is still useful, hence the merge.

PR #64, opened 2020-05-20
add cx_ecfp_scalar_mult syscall

greenknot:
Ed25519 is the only curve supported for now, but other curves should be easier to add because they are supported by OpenSSL.


src/cx_ec.c
```diff
@@ -882,3 +883,51 @@ int sys_cx_ecdh(const cx_ecfp_private_key_t *key, int mode, const uint8_t *publi
...
+      errx(1, "scalarmult_ed25519 failed");
+    }
+
+    BN_bn2binpad(Qx, P + 1, 32);
+    BN_bn2binpad(Qy, P + 33, 32);
+    break;
+  case CX_CURVE_SECP256K1:
+  case CX_CURVE_SECP256R1:
+  default:
+    errx(1, "TODO: unsupported curve");
```
niooss-ledger, line 922: Would it be possible to indicate in this message which curve was requested (ie. ``curve``)?

src/cx_ed25519.c
```diff
@@ -0,0 +1,190 @@
...
+    BN_free(d2);
+    BN_free(q);
+    BN_free(q_minus_two);
+    return -1;
+  }
+
+  one = BN_value_one();
+  BN_dec2bn(&two, "2");
+  BN_dec2bn(&q, constant_q);
+  BN_sub(q_minus_two, q, two);
```
niooss-ledger, line 154: Where is ``q_minus_two`` used?

src/cx_ed25519.c
```diff
@@ -0,0 +1,190 @@
...
+static BN_CTX *ctx;
+
+typedef struct {
+  BIGNUM *x;
+  BIGNUM *y;
+} POINT;
+
+static void inv(BIGNUM *r, BIGNUM *x)
+{
+  BN_mod_inverse(r, x, q, ctx);
```
niooss-ledger, line 22: It might make the code easier to read if this function is inlined in its callers

niooss-ledger:
I added some test vectors. The PR looks otherwise good to me, thanks :)

PR #66, opened 2020-05-22
cx_edward_compress

greenknot:


tests/syscalls/test_ec.c
```diff
@@ -122,9 +178,75 @@ void test_scalar_mult_ec25519(void **state __attribute__((unused))) {
...
+    assert_int_equal(pub.W[0], 4);
+    sys_cx_edward_compress_point(curve, (uint8_t *) &pub.W, pub.W_len);
+
+    assert_int_equal(pub.W[0], 2);
+    assert_memory_equal(expected_key, &pub.W[1], mpi_size);
+  }
+}
+
+static void test_ed25519_get_public_key(void **state) {
+  (void)state;
```
niooss-ledger, line 212: In another function of this file, there is an `unused` attribute instead: `void test_scalar_mult_ec25519(void **state __attribute__((unused))) {`.
It might be better to be consistent and use attributes everywhere or `(void)state;` everywhere, but not both.

src/cx_ed25519.c
```diff
@@ -180,3 +187,168 @@ int scalarmult_ed25519(BIGNUM *Qx, BIGNUM *Qy, BIGNUM *Px, BIGNUM *Py, BIGNUM *e
 
   return scalarmult(&Q, &P, e);
 }
+
+void cx_encode_int(uint8_t *v, size_t len)
```
niooss-ledger, line 191: More that "encode/decode", this is about endianness (converting Big Endian numbers from/to Little Endian, because Ed25519 "exported data" is Little Endian but the computations in Speculos are Big Endian). I suggest renaming these functions `be2le` and `le2be`, or at least adding comments that describe their purposes.

src/cx_ed25519.c
```diff
@@ -180,3 +187,168 @@ int scalarmult_ed25519(BIGNUM *Qx, BIGNUM *Qy, BIGNUM *Px, BIGNUM *Py, BIGNUM *e
...
+  if (a == NULL || xx == NULL) {
+    ret = -1;
+    goto free_bn;
+  }
+
+  BN_mul(a, d1, y, ctx);
+  BN_mul(a, a, y, ctx);
+  BN_add(a, a, one);
+  BN_mod_inverse(a, a, q, ctx);
+  BN_mul(xx, y, y, ctx);
```
niooss-ledger, line 267: This could be optimized, as ``y*y`` is computed twice (once in ``a = d1 * y * y``, and then here with ``xx = y * y``)

greenknot:
Thanks for the review!

PR #71, opened 2020-05-28
fix sys_cx_math_add with arithmetic overflow

greenknot:


src/emu_cx_math.c
```diff
@@ -87,13 +120,13 @@ int sys_cx_math_add(uint8_t *r, const uint8_t *a, const uint8_t *b, unsigned int
   BN_bin2bn(b, len, bb);
 
   BN_add(rr, aa, bb);
-  BN_bn2binpad(rr, r, len);
+  carry = BN_bn2binpad_truncate(rr, r, len);
```
niooss-ledger, line 123: This implicitly converts a `bool` to an `int`, assuming that `true` is `1`. This is right, according to C99 (cf. section 7.16 in http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1124.pdf):
> They are
> `true`
> which expands to the integer constant 1,
> `false`
> which expands to the integer constant 0

Nevertheless IMHO the code would be more readable with an explicit cast to `int` (or with `carry = BN_bn2binpad_truncate(rr, r, len) ? 1 : 0;`). What do you think?

greenknot: Indeed, let's go for the second option.

greenknot:
Merged using command line.

PR #72, opened 2020-06-02
add SLIP10 support

greenknot:


src/emu_os_bip32.c
```diff
@@ -315,6 +325,42 @@ static int hdw_bip32_ed25519(extended_private_key *key, const uint32_t *path, si
   return 0;
 }
 
+static int hdw_slip10(extended_private_key *key, const uint32_t *path, size_t length, uint8_t *private_key, uint8_t* chain)
+{
+  uint8_t tmp[1+64+4];
+  unsigned int i;
+
+  for (i = 0; i < length; i++) {
+    if (((path[i] >> 24) & 0x80) != 0) {
```
niooss-ledger, line 334: In order to check the most significant bit, `if ((path[i] >> 31) != 0)` or `if ((path[i] & 0x80000000) != 0)` seem more suited.

greenknot: I just added a `is_hardened_child` function.

PR #78, opened 2020-06-23
Fix lgtm.com warnings in vnc_server.c

niooss-ledger:


vnc/vnc_server.c
```diff
@@ -261,6 +261,7 @@ int main(int argc, char **argv)
     sigset_t mask;
     bool verbose;
     fd_set fds;
+    size_t frambuffer_size;
```
greenknot, line 264: There's a typo in the variable name (`frambuffer_size`)

niooss-ledger: Thanks! Fixed.

PR #81, opened 2020-06-23
Test syscalls cx_hash_sha256 and cx_hash_sha512

niooss-ledger:
The syscalls `cx_hash_sha256` and `cx_hash_sha512` perform `hash_init`, `hash_update` and `hash_final` in a single syscall, and return the size of the digest. Test their implementations.


greenknot:
Merged manually (001f9f36ad958e77d53ba3b3791bd2f4a8dca892).

PR #84, opened 2020-06-24
Proposal for screen state event automation

ogoldgr:
This PR modifies `automation_server.py` implementation so that display events
are accumulated within a list and broadcasted to clients at once.

This allows a simpler event processing at client side as it does not receive
uncorrelated events anymore but the complete list of the displayed items instead.

**Note**: bug found if ui swaps to another screen in the middle of a refresh
resulting in a list containing display items mixed from both screens.

### DIFF
--- a/mcu/automation_server.py
+++ b/mcu/automation_server.py
@@ -16,6 +16,8 @@ class AutomationServer(socketserver.ThreadingMixIn, socketserver.TCPServer):
 
         self.logger = logging.getLogger("automation")
         self.clients = []
+        self.state = []
+        self.prev_state = []
 
     def add_client(self, client):
         self.clients.append(client)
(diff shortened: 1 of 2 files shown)


gre:
**TLDR: the idea is interesting but I'm not sure if the PR studies all the tradeoffs it implies, we have some apps that might not have this well supported.**

**If this is not breaking changes from a speculos usage, we can always merge it.**

can you share some example of  what it means from user land point of view, like is this new event in the automation stream? (the thing that emits a series of `{ x, y, text }` ?

---

to give some context, we have stated to heavily use that paradigm in Ledger Live team to do transaction automatically and do a lot of tests with this ( https://github.com/LedgerHQ/ledger-live-common/blob/master/docs/bot.md ). What I can say from experience accumulated is that we can't always assume that a screen is static, sometimes there are "scrolling menu" like the the Cosmos app.

What we did on our side is we still consume the raw events but we did the abstraction on user land, with this current implementation: https://github.com/LedgerHQ/ledger-live-common/blob/master/src/bot/specs.js#L54
which consist of describing some sort of "spec" of all screens of an app like this:

https://github.com/LedgerHQ/ledger-live-common/blob/master/src/families/cosmos/speculos-deviceActions.js#L16-L112

It might be a bit early from my point of view to tell if our paradigm as well as this proposal is powerful enough to cover other devices like Blue (which is supported by speculos). At first glance, we saw in our side that Blue support is challenging to reconciliate, but not possible.

gre (commented):
(sent my review in a comment on this PR, commenting here just to clear my list of things to review 👍 )

greenknot:
@ogoldgr does the new [REST API](https://github.com/LedgerHQ/speculos/blob/9f67004cc5aa68d0e938f1071b17d7ac86c9b3b6/docs/user/api.md) answer your needs (`/events?stream=false`)?

greenknot:
> @ogoldgr does the new [REST API](https://github.com/LedgerHQ/speculos/blob/9f67004cc5aa68d0e938f1071b17d7ac86c9b3b6/docs/user/api.md) answer your needs (`/events?stream=false`)?

Closing, please reopen the issue if the API doesn't answer your needs.

PR #87, opened 2020-07-06
Make the code coverage deterministic

niooss-ledger:
* Force the seed when running the code coverage, using `RNG_SEED` environment varible.
* Use `RAND_set_rand_method()` to make `RAND_bytes()` get random bytes from the same PRNG as `sys_cx_rng()`

The aim of this Pull Request is to fix blinking in code coverage analysis such as https://codecov.io/gh/LedgerHQ/speculos/pull/86/changes

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -50,6 +50,7 @@ jobs:
     - name: Rebuild with code coverage instrumentation
       env:
         CTEST_OUTPUT_ON_FAILURE: 1
+        RNG_SEED: 0
       run: |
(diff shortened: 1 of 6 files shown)


gre:
nice!

TamtamHero:
That's excellent
Does it also mean that we're getting deterministic signature when using speculos with `--deterministic-rng` ?

niooss-ledger:
> Does it also mean that we're getting deterministic signatures when using speculos with `--deterministic-rng` ?

Yes. More precisely, I have tested with a small app that uses `cx_ecdsa_sign(privateKey, CX_RND_TRNG | CX_LAST, ...)` to sign the same message 3 times. The 3 signatures are different (this is not "deterministic signature" like RFC 6979), but they are always the same when running the app with the same RNG seed and the same requests (sent in the same order). This matches what I would expected from a deterministic PRNG.

PR #89, opened 2020-07-10
Fix eddsa verification

jleni:
Verifying compressed keys was always failing because keys composed of (1+32) bytes were rejected.


jleni:
There is something a bit confusing... with this fix I managed to get eddsa verification to work with my 32 bytes pubkey (+ 0x02 prefer)... However, the exact same code fails in a real device with a 0x6081 error code.
I wonder if this very same issue exist in the 1.6 firmware..

Issue #90, opened 2020-07-10
ECDSA - RFC6979 support broke signing with blake2b 


jleni:
The following commit https://github.com/LedgerHQ/speculos/commit/3708d0732c75bc2e18823d959386dff497d11a0d broke blake2b ecdsa signatures

This check is triggered:
https://github.com/LedgerHQ/speculos/blob/94adf3ea90612eed2288daf3befd6f1b3d9b9205/src/cx_rng_rfc6979.c#L164-L167

because blake2b
https://github.com/LedgerHQ/speculos/blob/4b531e8e777a5769b20042793d6297a5830b9257/src/emu_cx_hash.c#L107-L116

https://github.com/LedgerHQ/speculos/blob/4b531e8e777a5769b20042793d6297a5830b9257/src/cx_hash.h#L173-L184

Checking different hash types it seems that `output_size` is set to zero for variable size hash functions. 

The `cx_rng_rfc6979_init` received hash length `h1_len` so I think that could be used instead.
https://github.com/LedgerHQ/speculos/blob/94adf3ea90612eed2288daf3befd6f1b3d9b9205/src/cx_rng_rfc6979.c#L158-L162

PR #91, opened 2020-07-13
Do not use RFC6979 signature with variable-length hashes

niooss-ledger:
The RFC6979 deterministic signature implementation in Speculos is not compatible with hash functions that have a variable output length, because they rely on this function to compute HMAC:

    int cx_hmac_init(cx_hmac_ctx *ctx, cx_md_t hash_id, const uint8_t *key,
                     size_t key_len)

In order to support variable-output-size digests, the digest size has to be specified, for example with a new `size_t output_size` parameter, like what was done for `cx_hash_init_ex`.

Instead of changing the API, keep the behaviour of `cx_ecdsa_sign(CX_RND_RFC6979, CX_BLAKE2B)` failing with `THROW(INVALID_PARAMETER);` and fix `cx_ecdsa_sign(CX_RND_TRNG, CX_BLAKE2B)` to make it work again
(with a weak PRNG used to make tests deterministics).

Fixes: https://github.com/LedgerHQ/speculos/issues/90

### DIFF
--- a/src/cx_ec.c
+++ b/src/cx_ec.c
@@ -766,6 +766,7 @@ int sys_cx_ecdsa_sign(const cx_ecfp_private_key_t *key, int mode, cx_md_t hashID
   int nid = 0;
   uint8_t *buf_r, *buf_s;
   const BIGNUM *r, *s;
+  const cx_hash_info_t *hash_info;
 
   const cx_curve_domain_t *domain = cx_ecfp_get_domain(key->curve);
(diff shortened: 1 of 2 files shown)


defunctzombie:
What is the correct approach if the code we want to test with the emulator invokes `cx_ecdsa_sign(CX_RND_RFC6979, CX_BLAKE2B)`? It seems like this code will work on the device but not on the emulator?

PR #95, opened 2020-08-18
Fix misspelling of "Ed25519"

niooss-ledger:
"EC25519" does not make sense. The function is testing Ed25519, so name it as such.


greenknot:
Merged manually (a339ba8721ed4c58145806446b6bc917227ed48d).

PR #99, opened 2020-09-15
Added missing glyphs. Fixes #98.

### LINKED ISSUE #98: Some glyphs are missing
Not all glyphs seem to be supported by speculos.
Met this problem with "CHECK_BADGE" glyph, but probably a lot of more are missing.

kingofpayne:
Added CHECK_BADGE, CROSS_BADGE, TRANSACTION_BADGE and EYE_BADGE glyphs.


greenknot:
Thanks!

PR #103, opened 2020-09-23
Support for scalar multiplication on secp256k1 and secp256r1

jibeee:
Scalar multiplication was only supported for Ed25519. This PR adds (partial) support for secp256k1 and P-256 curves.


greenknot:
Rebased and merged (76f42ca86b8eea3c82a2d9f34e2af4c60b0e9831). Thanks!

PR #107, opened 2020-10-25
Added experimental recording to GIF

kingofpayne:
The new option --gif can be enabled to record the screen during
application execution, and save to an animated GIF file. This should be
great for demos!

### DIFF
--- a/mcu/bagl.py
+++ b/mcu/bagl.py
@@ -70,7 +70,7 @@ class Bagl:
         self.logger = logging.getLogger("bagl")
 
     def refresh(self):
-        self.m.update()
+        self.m.refresh()
 
     def hal_draw_bitmap_within_rect(self, x, y, width, height, colors, bpp, bitmap, restore=None):
         if bpp == 3 or bpp > 4:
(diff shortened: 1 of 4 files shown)


greenknot:
Thanks for the PR @kingofpayne and sorry for the lack of feedback.

I was quite reluctant to add another option to the command-line but I think the new API which is going to be merged could be used to do that:


```python
#!/usr/bin/env python3

import io
import requests
from PIL import Image

API_URL = "http://127.0.0.1:5000"

def take_screenshot(session):
    response = session.get(f"{API_URL}/screenshot")
    return response.content

def record_images():
    images = []

    try:
        with requests.session() as session:
            with session.get(f"{API_URL}/events", stream=True) as stream:
                while True:
                    image = take_screenshot(session)
                    if len(images) == 0 or image != images[-1]:
                        images.append(image)
                    line = stream.raw.readline()
                    print(f"got event: {line}")
                    if not line:
                        break
    except KeyboardInterrupt:
        pass

    return images

def save_gif(filename, images, duration=500):
    images = [ Image.open(io.BytesIO(image)) for image in images ]
    images[0].save(filename, save_all=True, append_images=images[1:], duration=duration, loop=0)

if __name__ == "__main__":
    images = record_images()
    save_gif("/tmp/speculos.gif", images)
```

greenknot:
Superseded by #195.

PR #114, opened 2020-12-11
Fix the cx_aes_iv syscall for SDK 1.6

aguinet:
The cx_aes_iv syscall number is indeed different in the SDK 1.6 than in
1.5 and Blue.

### DIFF
--- a/src/emulate_1.5.c
+++ b/src/emulate_1.5.c
@@ -4,6 +4,7 @@
 #include "emulate.h"
 
 #include "bolos_syscalls_1.5.h"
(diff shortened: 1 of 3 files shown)


greenknot:
Good catch. Can you also add it to `emulate_1.5.c` and `emulate_blue_2.2.5.c` please?

aguinet:
I hesitated doing it. Indeed, if I remember correctly, the syscall number only differs for the 1.6 SDK, and is the same for 1.5 and blue. But you tell me :)

greenknot:
Thanks!

PR #115, opened 2020-12-11
AES CBC decrypt: fix case when in & out alias

aguinet:
I still have to check whether this is a supported case on the actual device!


greenknot:
Thanks!

PR #116, opened 2020-12-14
Implement scalar multiplication on Curve25519

aguinet:
This patch implements scalar multiplication on Curve25519 for the `cx_ecfp_scalar_mult` syscall.

I am not completely found of the implementation, because it copy/past the OpenSSL implementation. The issue is that the publicly accessible `X25519` function isn't a "clean" scalar multiplication per se, because it applies the "DJB" recommendations on the scalar bytes (https://github.com/openssl/openssl/blob/master/crypto/ec/curve25519.c#L4384). Please let me know if you see a "cleaner" way to do this :)

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,6 +1,7 @@
 add_library(emu
         cx_aes.c
         cx_blake2b.c
+        cx_curve25519.c
         cx_crc.c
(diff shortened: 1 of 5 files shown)


greenknot:
Thanks @aguinet!

PR #117, opened 2020-12-18
add support for the Nano X

greenknot:


doc/usage.md
```diff
@@ -6,24 +6,27 @@
 ./speculos.py apps/btc.elf
 ```
 
-The Ledger Nano S is the default model and the Ledger Blue can be specified on
-the command-line:
+The Nano S is the default model; the Nano X and Blue can be specified on the
+command-line:
 
 ```console
+./speculos.py --model nanox apps/btc-blue.elf
```
niooss-ledger, line 13: It seems that the file name needs to be changed too (`btc-nanox.elf`?)

src/emu_seproxyhal.c
```diff
@@ -12,27 +12,26 @@ static bool status_sent = false;
...
       if (errno == EINTR)
         continue;
       warn("read from seph fd failed");
       return -1;
     }
+    count -= n;
   }
 
-  return n;
+  return count;
```
niooss-ledger, line 34: Returning `count`, which is decremented, is strange: now `readall` returns the number of bytes which were *not* read. And later in `sys_io_seproxyhal_spi_recv` the returned size is not really used, which is normal (as the received size is the size given in `readall` call).

Maybe change the result as "error status" and use `return 0` in `readall` ?

PR #120, opened 2020-12-31
Package Speculos to be installed with pip

grydz:


.github/workflows/continuous-integration-workflow.yml
```diff
@@ -26,14 +26,14 @@ jobs:
 
     - name: Build
       run: |
-        cmake -Bbuild -H. -DPRECOMPILED_DEPENDENCIES_DIR=/install -DWITH_VNC=1
```
greenknot, line 29: Is it necessary to rename the `build` folder? What does it fix?

CMakeLists.txt
```diff
@@ -7,7 +7,11 @@ include(ExternalProject)
 set(CMAKE_C_COMPILER arm-linux-gnueabihf-gcc)
 set(CMAKE_EXE_LINKER_FLAGS -static)
 
-enable_testing()
```
greenknot, line 10: I wonder if `BUILD_TESTING` could be used instead of `WITH_TEST` (https://cmake.org/cmake/help/v3.15/command/enable_testing.html)?

speculos.py
```diff
@@ -28,6 +28,8 @@
 from mcu.finger_tcp import FakeFinger
 from mcu.vnc import VNC
 
+LOG = logging.getLogger("speculos")
```
greenknot, line 31: Please keep `logger` in lowercase

greenknot (changes requested):
Nice!

.github/workflows/continuous-integration-workflow.yml
grydz: Unfortunately, `setuptools` also builds the Python package in a directory named `build`. It's a mess if you do `pip install .` in speculos folder...

speculos.py
grydz: Fixed in 5f43cc1225b3dd34cc47aa1e0013b8b6badaa938.

CMakeLists.txt
grydz: It seems to work! See 5f43cc1225b3dd34cc47aa1e0013b8b6badaa938.

.github/workflows/continuous-integration-workflow.yml
greenknot: I'd prefer to keep speculos build in the `build` directory and find a workaround for `setuptools` (eg. `--dist-dir` or something like this?)

niooss-ledger:
Hello, as I am also interested in making `pip install https://github.com/LedgerHQ/speculos` work, I took a look at this Pull Request and hit several issues. So I am writing down here the issues I faced and the way I think they could be solved.

First, Speculos currently documents using the `build/` directory to build the C code (with `cmake`). This conflicts with setuptools, but this can be fixed by creating a `setup.cfg` containing:
```ini
[build_py]
build_lib=build-pythonlib

[install_lib]
build_dir=build-pythonlib
```

Then, `setup.py` packages the whole `build` directory, even tough only `build/src/launcher` is needed (this directory also contains `cmake` artefacts, compiled objects, etc.). I did not find an elegant way of solving this issue and I am thinking of copying this file in some other place which is included in the Python package which is built.

Another issue is that `find_packages()` in `setup.py` returns `["mcu", "mcu.resources"]`. So the `api` directory is missing (and this is because it does not contain a `__init__.py` file) but more importantly, files are directly installed in `site-packages` (for example in `/usr/local/lib/python3.8/site-packages/mcu` instead of `/usr/local/lib/python3.8/site-packages/speculos/mcu`). It seems most Python project do not have this issue by using a subdirectory to store the Python modules. Here, it is possible to move `api`, `mcu` and `speculos.py` in a new `speculos/` directory, but doing so breaks the logic used to find the launcher, in https://github.com/LedgerHQ/speculos/blob/4d1aed38b2f88505258fe1742995df6e4839ceba/speculos.py#L35
Maybe there exists a way to tell setuptools that the current directory is the root of the `speculos` module?

TL;DR: before being able to package Speculos as a Python package (which would make running `pip install https://github.com/LedgerHQ/speculos` work), a new way of "installing" the launcher needs to be defined. And then some other issues will still need to be solved.

niooss-ledger:
This work has been continued and integrated in Pull Requests https://github.com/LedgerHQ/speculos/pull/201 (to introduce a `main` function), https://github.com/LedgerHQ/speculos/pull/212 (to add a `setup.py` file), https://github.com/LedgerHQ/speculos/pull/219 (to avoid building tests in the Python package) and https://github.com/LedgerHQ/speculos/pull/220 (to make the CI publish the package).

Therefore this Pull Request can be closed.

PR #122, opened 2021-01-13
Fix invisible window when using multiple screens.

dmorais-ledger:
This commit fix a problem occurring when using multiple screens, with different resolutions, and with current screen placed in such a way that default _absolute coordinate_ of 10,10 is in a non visible area...
Now, on first launch, the default coordinates used takes in account current screen geometry, and check for each launch that window will be fully visible in one of the available screens : if it is clipped, even a little bit, default coordinates (in current screen) will be used.


greenknot:
Thanks!

PR #123, opened 2021-01-15
Fix issue when selecting Quit in App with down key

dmorais-ledger:
When you select Quit in the App with the down key, the program was doing a sys.exit(1) which let the terminal (in text mode) in an improper state, and does not save window position when using qt display.

### DIFF
--- a/mcu/headless.py
+++ b/mcu/headless.py
@@ -2,6 +2,7 @@ import select
 
 from . import bagl
 from .display import Display, FrameBuffer, MODELS, RENDER_METHOD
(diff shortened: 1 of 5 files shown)


greenknot (changes requested):
Thanks for the PR! Just a small feedback: can you introduce a new exception (eg. `ReadError`) instead of the built-in exception `RuntimeError`?

dmorais-ledger:
You're right, using a dedicated exception is much cleaner: thanks for the suggestion!

greenknot (approved):

PR #124, opened 2021-01-26
fixed wrong interpretation of 16bit signed coordinates

shufps:
In layouts like `nnbnn` normally the first and the last lines are displayed half. The other half is outside of the visible screen.

On the simulator, the first line wasn't drawn at all. The reason for this is that the y-coordinate of the bagl image is actually 16bit signed (may be negative) but is interpretated as 32bit unsigned on the simulator.

A coordinate of y=-5 becomes y=65531 which is outside the screen, so it's not drawn at all.

This fix casts the 32bit unsigned in the right value and the missing line is displayed again.

### DIFF
--- a/mcu/bagl.py
+++ b/mcu/bagl.py
@@ -576,8 +576,8 @@ class Bagl:
 
     def display_raw_status(self, data):
         if data[0] == SEPROXYHAL_TAG_SCREEN_DISPLAY_RAW_STATUS_START:
-            x = int.from_bytes(data[1:3], byteorder='big')
-            y = int.from_bytes(data[3:5], byteorder='big')
+            x = int.from_bytes(data[1:3], byteorder='big', signed=True)
+            y = int.from_bytes(data[3:5], byteorder='big', signed=True)
             w = int.from_bytes(data[5:7], byteorder='big')
             h = int.from_bytes(data[7:9], byteorder='big')
             bpp = int.from_bytes(data[9:10], byteorder='big')


greenknot:
Nice catch!

Do you think the following patch fixes the issue (it looks a bit closer to the [SDK](https://github.com/LedgerHQ/nanos-secure-sdk/blob/1bc941792868aadb29e45c771ccf997b9186b305/include/bagl.h#L68-L69) but I'm not sure about it)?

```diff
diff --git a/mcu/bagl.py b/mcu/bagl.py
index 81b7490..5348836 100644
--- a/mcu/bagl.py
+++ b/mcu/bagl.py
@@ -9,8 +9,8 @@ from . import bagl_glyph
 bagl_component_t = Aligned(4, Struct(
     "type"    / Int8ul,
     "userid"  / Int8ul,
-    "x"       / Int16ul,
-    "y"       / Int16ul,
+    "x"       / Int16sl,
+    "y"       / Int16sl,
     "width"   / Int16ul,
     "height"  / Int16ul,
     "stroke"  / Int8ul,
```

shufps:
Oh nice - probably that's the way how to do it right :sweat_smile: 

I'll test it asap :)

shufps:
No sorry - that fix you suggested doesn't fix the problem :man_shrugging:

shufps:
that's strange ...
```python
from construct import *

bagl_component_t = Aligned(4, Struct(
    "x"       / Int16sl,
    "y"       / Int16sl,
))

component = bagl_component_t.parse(b'\xff\xff\xff\xff')
print("%d" % component.x)
```
correctly outputs `-1` ... so it must be working ...

I'll try to figure it out ...

shufps:
Ah this is the conversion that messes it up:
https://github.com/LedgerHQ/speculos/blob/4ff4f35f1868197caa1141f3ec249f17ffaa200c/mcu/bagl.py#L587

Do you have any ideas how to cleanly fix it? :)

greenknot:
Maybe with the additional parameter `signed=True` (https://docs.python.org/3/library/stdtypes.html#int.from_bytes)?

shufps:
@greenknot thx, that's working :star_struck: 

I updated the PR :)

greenknot:
Great, thanks!

PR #125, opened 2021-01-28
reorganize source code

greenknot:
The current source code architecture is messy.

Move everything related to syscall implementations to `bolos/` and move half of `launcher.c` to `svc.c`.


src/bolos/cx_aes.c
```diff
@@ -1,13 +1,13 @@
-#include <assert.h>
 #include <err.h>
+#include <assert.h>
```
niooss-ledger, line 2: Why did you change the order of the headers? `clang-format` sorts the headers alphabetically, by default. Did you use another option to sort them differently?

src/svc.c
```diff
@@ -0,0 +1,283 @@
+#define _GNU_SOURCE
```
niooss-ledger, line 1: To better know why this is needed, I suggest adding a comment:
```c
#define _GNU_SOURCE /* for memmem() */
```

src/sdk.h
```diff
@@ -0,0 +1,11 @@
+#pragma once
+
+typedef enum {
+  SDK_NANO_X_1_2,
+  SDK_NANO_S_1_5,
+  SDK_NANO_S_1_6,
+  SDK_BLUE_2_2_5,
+  SDK_LAST
```
niooss-ledger, line 8: `SDK_LAST` is misleading, as this value is "invalid", or the "count"/"numbers" of versions.

This PR did not introduce `SDK_LAST` (it was previously in `src/emulate.h`), but this PR could be the opportunity to rename this value to something such as `SDK_COUNT` (to indicate that this value is both the last one of `sdk_version_t` so it is suitable in a `for` loop, and that this value is not valid, because it is "one above the last valid version").

src/svc.h
```diff
@@ -0,0 +1,11 @@
+#pragma once
+
+#include <signal.h>
```
niooss-ledger, line 3: This file used types that require more `#include`: `bool`, `size_t`... Please add:
```c
#include <stdbool.h>
#include <stddef.h>
```

niooss-ledger (commented):
This looks good :+1:. I commented on some minor things that should be changed before merging, IMHO.

src/svc.h
niooss-ledger: ... I have directly done this.

src/bolos/cx_aes.c
greenknot: No reason...

src/sdk.h
greenknot: Done

greenknot:
Thanks @niooss-ledger! I force-pushed the suggested changes.

greenknot:
Rebased and merged manually: 9862118edcc3deb0fa8149e210edc87336a2b008.

PR #126, opened 2021-01-29
Add codecov.io and lgtm.com badges to README

niooss-ledger:
This helps following the automated code analysis.


greenknot:
Oh noes! Does it mean I'll have to fix these warnings and add more tests? :fearful:

PR #132, opened 2021-02-05
improve argument passing using namedtuple

greenknot:


greenknot:
This PR doesn't introduce any functional changes, but should improve readability.

I never used Python type annotations before so feedback is welcome.

dkremer-ledger (commented):
Should be `NoneType` instead of `None`, but in general if it really it returns `None`, I think you can leave the typing for the return value.

dkremer-ledger (approved):

PR #136, opened 2021-02-11
Fix install params mapping

Saltari:
This PR aimed at:
- ignoring core files in versioning in case of a crash
- replacing `getpagesize` by `sysconf(_SC_PAGESIZE)` which is preferred as part of posix
- mapping an extra page for the app's install parameters :
The install parameters of an app are mapped at the end of the `nvram` data of an app.
If this address happens to be the end of a page, speculos crashes when trying to access these parameters
ex: 
- `page_size = 0x1000`, elf size is `0x23d00` which means it will be mapped between `[0x40000000 - 0x40024000]` => as `_install_parameters` are mapped at the end of `_nvram_data` so `0x400023d00` they can be accessed ✔️ 
- `page_size = 0x1000`, elf size is `0x24000` which means it will be mapped between `[0x40000000 - 0x40024000]` => as `_install_parameters` are mapped at the end of `_nvram_data` so `0x400024000` they cannot be accessed ⛔ 
Mapping an additional page helps with this last case

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@
 __pycache__/
 apps/*vault*
 build/
+core
--- a/src/launcher.c
+++ b/src/launcher.c
@@ -68,12 +68,12 @@ static struct app_s *current_app;
 
 static void *get_lower_page_aligned_addr(uintptr_t vaddr)
 {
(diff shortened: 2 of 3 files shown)


greenknot:
Thanks!

PR #138, opened 2021-03-01
[tests] improve 'app' fixture

dkremer-ledger:
The app fixture was requiring an extrafeature to be stopped.
In case of failure, the APDUClient was not properly killed,
thus leading to cascading test cases. This commit aims to
fix that.

### DIFF
--- a/tests/apps/conftest.py
+++ b/tests/apps/conftest.py
@@ -52,13 +52,13 @@ def list_apps_to_test(app_dir) -> List[AppInfo]:
                 device=matching.group(1),
                 name=matching.group(2),
                 version=matching.group(3),
-                hash=matching.group(4)
+                hash=matching.group(4),
             )
         )
     return all_apps
 
 
-@pytest.fixture(scope='function')
+@pytest.fixture(scope="function")
 def app(request):
     _app = APDUClient(request.param)
     yield _app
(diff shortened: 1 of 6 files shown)


tests/apps/test_btc.py
```diff
@@ -8,57 +8,30 @@
...
+
+
+from .utils import Vnc
+
 
 class TestBtc:
-    '''Tests for Bitcoin app.'''
+    """Tests for Bitcoin app."""
+
+    app_names = ['btc']
```
dkremer-ledger, line 19: app names attached in test class instead of specific filter function

tests/apps/conftest.py
```diff
@@ -26,42 +23,63 @@ def finger_client():
...
+    """
+    # name example: nanos#btc#1.5#5b6693b8.elf
+    app_regexp = re.compile(
+        r"(nanos|nanox|blue)#(.+)#([^#][\d\w\-.]+)#([a-f0-9]{8}).elf"
+    )
+    all_apps = []
+    for filename in os.listdir(app_dir):
+        matching = re.match(app_regexp, filename)
+        if not matching or len(matching.groups()) != 4:
+            continue
```
dkremer-ledger, line 47: filter app names for a specific pattern

PR #141, opened 2021-03-10
LVFM-846: add missing syscall to Speculos' Blue emulation

hcleonis:
A Blue device with firmware 2.2.10-ee(l)(d) allow to call the `os_perso_derive_node_bip32_seed_key()` and `cx_hmac_sha512()` Bolos syscalls. However it is not possible for a Blue app running in Speculos, although the code for these syscalls is emulated for LNS/LNX. The syscall interface to these emulated functions is missing in Speculos' Blue emulation. 

This PR adds the appropriate syscall interfaces to these 2 Bolos emulated functions.
All tests green.

### DIFF
--- a/src/bolos/cx_hmac.h
+++ b/src/bolos/cx_hmac.h
@@ -1,10 +1,32 @@
 #pragma once
 
+/**
+ * HMAC context.
+ */
+typedef struct {
+  cx_hash_for_hmac_ctx hash_ctx;
+  uint8_t key[128];
+} cx_hmac_ctx;
+
+/* Aliases for compatibility with the old SDK */
+typedef cx_hmac_ctx cx_hmac_t;
+typedef cx_hmac_ctx cx_hmac_ripemd160_t;
+typedef cx_hmac_ctx cx_hmac_sha256_t;
+typedef cx_hmac_ctx cx_hmac_sha512_t;
+
+void cx_scc_struct_check_hashmac(const cx_hmac_t *hmac);
+int cx_hmac_sha256(const unsigned char *key, unsigned int key_len,
+                   const unsigned char *in, unsigned int len,
+                   unsigned char *out, unsigned int out_len);
+
 int cx_hmac_init(cx_hmac_ctx *ctx, cx_md_t hash_id, const uint8_t *key,
                  size_t key_len);
 int cx_hmac_update(cx_hmac_ctx *ctx, const uint8_t *data, size_t data_len);
 int cx_hmac_final(cx_hmac_ctx *ctx, uint8_t *out, size_t *out_len);
 
+int cx_hmac_sha512(const unsigned char *key, unsigned int key_len,
+                   const unsigned char *in, unsigned int len,
+                   unsigned char *out, unsigned int out_len);
 int cx_hmac_sha512_init(cx_hmac_sha512_t *hmac, const unsigned char *key,
                         unsigned int key_len);
 
@@ -12,3 +34,6 @@ int sys_cx_hmac(cx_hmac_t *hmac, int mode, const unsigned char *in,
                 unsigned int len, unsigned char *out, unsigned int out_len);
 int sys_cx_hmac_sha256_init(cx_hmac_sha256_t *hmac, const unsigned char *key,
                             unsigned int key_len);
+
+#define sys_cx_hmac_sha256 cx_hmac_sha256
+#define sys_cx_hmac_sha512 cx_hmac_sha512
--- a/src/bolos/cx_scc.c
+++ b/src/bolos/cx_scc.c
@@ -2,6 +2,7 @@
 
 #include "bolos/exception.h"
 #include "cx.h"
+#include "cx_hmac.h"
 
 /* ======================================================================== */
 /* ===  MISC  === MISC === MISC === MISC ===  MISC  === MISC ===  MISC  === */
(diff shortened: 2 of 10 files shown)


src/bolos/cx_hmac.h
```diff
@@ -5,6 +5,9 @@ int cx_hmac_init(cx_hmac_ctx *ctx, cx_md_t hash_id, const uint8_t *key,
 int cx_hmac_update(cx_hmac_ctx *ctx, const uint8_t *data, size_t data_len);
 int cx_hmac_final(cx_hmac_ctx *ctx, uint8_t *out, size_t *out_len);
 
+int cx_hmac_sha512(const unsigned char *key, unsigned int key_len,
```
greenknot, line 8: Please remove `cx_hmac_sha512` from `cx.h` (and make appropriate changes such as including `cx_hmac.h` when necessary) in order to prevent this prototype from being declared twice.

src/emulate_blue_2.2.5.c
```diff
@@ -1,10 +1,18 @@
 #include <err.h>
 #include <stdio.h>
 
+#include "bolos/cx.h"
 #include "bolos/cx_aes.h"
+#include "bolos/cx_hmac.h"
 #include "bolos_syscalls_blue_2.2.5.h"
 #include "emulate.h"
 
+unsigned long sys_os_perso_derive_node_bip32_seed_key(
```
greenknot, line 10: This prototype is already declared in `emulate_1.2.c`, but it isn't really appropriate. Please move these prototypes to `bolos/os_bip32.h`, that looks better to me.

src/emulate_blue_2.2.5.c
```diff
@@ -60,6 +68,24 @@ int emulate_blue_2_2_5(unsigned long syscall, unsigned long *parameters,
...
+           unsigned int,         mode,
+           cx_curve_t,           curve,
+           const unsigned int *, path,
+           unsigned int,         pathLength,
+           unsigned char *,      privateKey,
+           unsigned char *,      chain,
+           unsigned char *,      seed_key,
+           unsigned int,         seed_key_length);
+  
+  SYSCALL6(cx_hmac_sha512, "(%p, %u, %p, %u, %p, %u)",
```
greenknot, line 82: The `cx_hmac_sha512` syscall ID is the same for each SDK version (ie. `0x600019cf`) and isn't specific to Blue 2.2.5. It can thus be moved to `src/emulate.c`

greenknot (changes requested):

src/bolos/cx_hmac.h
hcleonis: Looking more closely at the content of files `cx.h`, `cx_hmac.h` and `cx_hash.h`, I can see that there are HMAC-related definitions in all 3 of them. So shouldn't the best move be:
1. moving all the HMAC-related type definitions and function declarations from `cx.h` to `cx_hmac.h`
2. moving the #defines for `sys_cx_hmac_sha256` and `sys_cx_hmac_sha512` from `cx_hash.h` to `cx_hmac.h`
3. Adjusting the #include directives in the source files as necessary.

WDYT?

greenknot: LGTM 👍

hcleonis: Done

src/emulate_blue_2.2.5.c
hcleonis: This also required to move the aliased declaration of `sys_os_perso_derive_node_bip32_seed_key()` to `bolos/os_bip32.c` and to add a non-aliased declaration of that function in `emulate.h` (where all sys_os_xxx() functions are declared) for the build to succeed. I verified that both `sys_os_perso_derive_node_bip32_seed_key()` and `sys_os_perso_derive_node_with_seed_key()` symbols exist only once in the built binary and that they refer to the same address:
```
(hsmenv) ~/work/dev/speculos ~ nm build/src/launcher | grep -E 'sys_os_perso_derive_node_.+_seed_key' 
000202a9 W sys_os_perso_derive_node_bip32_seed_key
000202a9 T sys_os_perso_derive_node_with_seed_key
```

src/bolos/cx_scc.c
```diff
@@ -2,6 +2,8 @@
 
 #include "bolos/exception.h"
 #include "cx.h"
+#include "cx_hmac.h"
+
```
greenknot, line 6: This blank line makes the CI [fail](https://github.com/LedgerHQ/speculos/pull/141/checks?check_run_id=2079398603), please delete it.

src/emulate.c
```diff
@@ -374,6 +374,24 @@ int emulate_common(unsigned long syscall, unsigned long *parameters,
   SYSCALL2(os_endorsement_get_public_key_certificate, "(%d, %p)",
            unsigned char,   index,
            unsigned char *, buffer);
+
```
greenknot, line 377: There are 2 trailing whitespaces on this line, please remove them and it should be good :)

greenknot (changes requested):

src/bolos/cx_scc.c
hcleonis: done

src/emulate.c
hcleonis: done

greenknot:
Squashed, rebased and merged. Thanks @hcleonis!

PR #166, opened 2021-05-07
Fix cx mpi rng

dmorais-ledger:
This PR fix an issue where **cx_mpi_rng** (called by syscall **cx_bn_rng**) was returning a value from 0 to n-1, while the expected behaviour is to return a value from 1 to n-1.


docs/user/usage.md
```diff
@@ -30,6 +30,7 @@ against the SDK `1.5` on the Nano S:
 ```
 
 Supported SDK values for the `-k`/`--sdk` argument are:
+
```
greenknot, line 33: Please rebase on `master`

src/bolos/cx_mpi.c
```diff
@@ -984,12 +984,23 @@ cx_err_t cx_mpi_rand(cx_mpi_t *x)
 cx_err_t cx_mpi_rng(cx_mpi_t *r, const cx_mpi_t *n)
 {
   cx_err_t error;
+  cx_mpi_t *tmp;
```
greenknot, line 987: Instead of introducing a temporary BN, it might be easier to call `BN_rand_range` in a loop until the result is different than zero.

greenknot (changes requested):
Please also add non-regression tests for this syscall (you might look at `tests/syscalls/test_math.c` as an example).

docs/user/usage.md
dmorais-ledger: Done.

src/bolos/cx_mpi.c
dmorais-ledger: That's smart and avoid loosing _resolution_ in the random range: I like it; thanks for the suggestion ;-)

afsym (approved):

tests/syscalls/test_mpi_rng.c
```diff
@@ -0,0 +1,112 @@
...
+  ret = BN_set_word(n, 2);
+  assert_int_equal(ret, 1);
+  for (i = 0; i < 256; i++) {
+    error = cx_mpi_rng(r, n);
+    assert_int_equal(error, CX_OK);
+    value = BN_get_word(r);
+    assert_int_equal(value, 1);
+  }
+
+  // Loop several times and check that if n=3 r value was 1 & 2:
```
greenknot, line 55: I'm not sure if the following tests should be kept (these tests actually check that the values returned by `BN_rand_range` are valid.)

tests/syscalls/test_mpi_rng.c
```diff
@@ -0,0 +1,112 @@
...
+  error = cx_mpi_rng(r, n);
+  assert_int_equal(error, CX_INVALID_PARAMETER);
+
+  // Check that cx_mpi_rng returns CX_OK and r=1 if n=2:
+  ret = BN_set_word(n, 2);
+  assert_int_equal(ret, 1);
+  for (i = 0; i < 256; i++) {
+    error = cx_mpi_rng(r, n);
+    assert_int_equal(error, CX_OK);
+    value = BN_get_word(r);
```
greenknot, line 51: `BN_is_one`?

greenknot (changes requested):
Thanks for the tests! I made a few comments :)

tests/syscalls/test_mpi_rng.c
```diff
@@ -0,0 +1,112 @@
...
+  // Those tests will check that cx_mpi_rng is working as expected:
+  // cx_mpi_rng(r,n) => Generate a random value r in the range 0 < r < n.
+
+  r = BN_new();
+  assert_non_null(r);
+  n = BN_new();
+  assert_non_null(n);
+
+  // Check that cx_mpi_rng returns CX_INVALID_PARAMETER if n=0:
+  ret = BN_set_word(n, 0);
```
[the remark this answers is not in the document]
dmorais-ledger: Ok.Done.

tests/syscalls/test_mpi_rng.c
```diff
@@ -0,0 +1,112 @@
...
+  assert_int_equal(ret, 1);
+  error = cx_mpi_rng(r, n);
+  assert_int_equal(error, CX_INVALID_PARAMETER);
+
+  // Check that cx_mpi_rng returns CX_INVALID_PARAMETER if n=1:
+  ret = BN_set_word(n, 1);
```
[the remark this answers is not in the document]
dmorais-ledger: Ok. Done.

tests/syscalls/test_mpi_rng.c
dmorais-ledger: I had a bug in _cx_mpi_rng()_ where calling it with r=3 was always returning 1 (or 2, I don't remember).
That bug is now fixed, but it is maybe good to keep the test in case some future change in _cx_mpi_rng()_ will make that bug appear again...

tests/syscalls/test_mpi_rng.c
dmorais-ledger: I don't understand what you are suggesting, here: could you please provide the full code of the suggestion?

dmorais-ledger: I wasn't looking in the right place, sorry.
I understood what you suggested and I will do it ;-)

tests/syscalls/test_mpi_rng.c
greenknot: Let's say that if a similar bug is introduced later, it will be catched thanks to code review. I'm not sure unit tests are specially adapted to test randomness (related discussion [here](https://softwareengineering.stackexchange.com/questions/147134/how-should-i-test-randomness) and [there](https://stackoverflow.com/questions/32458/random-data-in-unit-tests)).

dmorais-ledger: Ok.

greenknot (approved):

PR #167, opened 2021-05-21
Added `--xy` option to customize window positioning from command-line.

ogoldgr:
Use case is to start multiple `speculos` instances:
- Instances are not started in sequence but in parallel.
- In that case we cannot rely on QT settings for that customization.


greenknot (approved):
Sorry for the late review, LGTM!

Issue #121, opened 2021-01-04
Automation HTTP service


aguinet:
It could be interesting to have Speculos creating an HTTP service that would serve an API which would allow to control automation (and the device itself) while it's running. This would potentially replace also the features provided by `--automation-port`, `--button-port` and `--finger-port`.

One use case I have in mind is being able to test Ledger companion apps on Windows & OSX using e.g. GitHub actions. Indeed, in GitHub actions, for a job, you can run a "background" service under Linux, that would be used by the job itself running on another OS. So, we could have Speculos running under Linux, exposing its automation port, for which we could push "automation JSON" depending on the test case.

This is a first "brain dump", and IMHO needs more thinking, especially on the HTTP API definition.

greenknot:
@aguinet done in #170, enjoy!

aguinet:
Wow this is very cool! Will test this ASAP :) Thanks @greenknot !

PR #170, opened 2021-06-10
add a REST API server

greenknot:
This PR adds a REST API server (new dependencies: `python3-flask` and `python3-flask-restful`). The server is disabled by default and can be enabled with the `--api` argument (`./speculos.py --api ./apps/btc.elf`).

The main reason for this PR is that there are way too many command-line arguments and it's difficult to interact with speculos using sockets (TCP ports have to be forwarded and custom clients have to be developed to interact with the finger/automation/button features). I hope that a REST API would make these features more easy to use. The issue #121 adds another use for this REST API server (updating automation rules on the fly). The existence of [Zemu](https://github.com/Zondax/zemu) also shows that a few features (screenshots?) were lacking.

The following URLs are available for now:

- http://127.0.0.1:5000/automation: replaces `--automation`
- http://127.0.0.1:5000/button/left, http://127.0.0.1:5000/button/right: replaces `--button-port` and emulates button press/release
- http://127.0.0.1:5000/events: replaces `--automation-port` and outputs text events
- http://127.0.0.1:5000/finger: replaces `--finger-port` and emulates finger touches
- http://127.0.0.1:5000/screenshot: takes a screenshot of the device screen

Before getting this PR merged, I would like to:

- Improve the API documentation (ideally generated and using swagger)
- Deprecate `--automation`, `--automation-port`, `--finger-port` and `--button-port`
- Keep Flask but maybe use other an other API library instead of flask-restful
- Update `Pipfile` and the port in Dockerfiles, etc.

But more important, please give us your feedback. Thanks!


greenknot:
The API can be previewed [here](https://petstore.swagger.io/?url=https://raw.githubusercontent.com/LedgerHQ/speculos/api/api/swagger/swagger.json).

niooss-ledger:
This looks great! I am starting to take a look and when I found [`api/swagger/`](https://github.com/LedgerHQ/speculos/tree/api/api/swagger) I wondered whether this directory was generated, and how it could be managed/updated. Later I found out that this directory was indeed generated, in a process documented in [`api/README.md`](https://github.com/LedgerHQ/speculos/blob/api/api/README.md).

Suggestion: could you add a file in [`api/swagger/`](https://github.com/LedgerHQ/speculos/tree/api/api/swagger) which states that this directory was automatically generated using instructions from `../README.md`?

api/api.py
```diff
@@ -0,0 +1,188 @@
...
+
+    def post(self):
+        args = self.parser.parse_args()
+        button = request.base_url.split("/")[-1]
+        if button == "both":
+            seph.handle_button(1, True)
+            seph.handle_button(2, True)
+            time.sleep(args.delay)
+            seph.handle_button(1, False)
+            seph.handle_button(2, False)
```
niooss-ledger, line 90: This feels like a `if action is None` / `... actions[action]` is missing from the case `button == "both"`.

api/swagger.yaml
```diff
@@ -0,0 +1,154 @@
...
+    get:
+      summary: "Get the events produced by the app"
+      parameters:
+      - name: "stream"
+        description: "Stream the events instead of returning a list"
+        in: query
+        required: false
+        schema:
+          type: boolean
+          default: true
```
niooss-ledger, line 74: Using `stream=true` by default makes the Swagger web UI "freeze" when issuing a call to the `/events` endpoint. Using `false` by default in the web UI seems more user-friendly.

speculos.py
```diff
@@ -236,16 +238,25 @@ def setup_logging(args):
         }
         args.sdk = default_sdk.get(args.model)
 
+    api_enabled = True
+    if args.api_port == 0:
+        api_enabled = False
```
niooss-ledger, line 243: Suggestion: use `api_enabled = (args.api_port != 0)` instead?

api/api.py
```diff
@@ -0,0 +1,188 @@
...
+        x = args.get("x")
+        y = args.get("y")
+        action = args.get("action", None)
+
+        if action is None:
+            seph.handle_finger(x, y, True)
+            time.sleep(args.delay)
+            seph.handle_finger(x, y, False)
+        else:
+            seph.handle_finger(x, y, actions[action])
```
niooss-ledger, line 169: `actions` is not defined here, so calling this API leads to:

    NameError: name 'actions' is not defined

niooss-ledger (changes requested):
Here are some comments on this PR. I found a bug in the finger API, which needs to be fixed. Otherwise this seems great!

api/api.py
greenknot: `action` is required now

api/api.py
greenknot: done

api/swagger.yaml
greenknot: done

speculos.py
```diff
@@ -236,16 +238,25 @@ def setup_logging(args):
...
     automation_path = None
     if args.automation:
+        logger.warn("--automation is deprecated, please use the REST API instead")
         automation_path = automation.Automation(args.automation)
 
     automation_server = None
     if args.automation_port:
+        logger.warn("--automation-port is deprecated, please use the REST API instead")
+        logger.warn("--automation-port is incompatible with the the API server, disabling it")
+        api_enabled = False
```
niooss-ledger, line 252: It seems strange to have two warnings. Is a `if api_enabled:` missing here?

greenknot: Actually it was intended, but let's add this condition

niooss-ledger (approved):

api/api.py
```diff
@@ -0,0 +1,188 @@
...
+from PIL import Image
+import pkg_resources
+import threading
+import time
+
+import mcu.automation
+
+static_folder = pkg_resources.resource_filename(__name__, "/swagger")
+
+app = Flask(__name__, static_url_path="/", static_folder=static_folder)
```
kingofpayne, line 15: On my side, app does not work in browser by default. setting `static_url_path=""` solves the issue.
Can you try ? (don't forget to empty browser cache as resources may be cached)

greenknot: Super weird, as discussed offline I can't reproduce the issue (with the same flask version, 0.12.2). Anyway, I applied your patch, thanks.

PR #172, opened 2021-06-15
Dockerfile: specify the name of the registry used for base images

niooss-ledger:
When using podman 3.1.2 to build Speculos containers, the command prompts the user for a container registry:

    $ podman build -f build.Dockerfile -t ghcr.io/ledgerhq/speculos-builder .
    STEP 1: FROM python:3.9-slim
      ▸ docker.io/library/python:3.9-slim
        quay.io/python:3.9-slim

Avoid this prompt by directly specifying the full name of the image, including the container registry.

While at it, sync `Dockerfile` to Python 3.9 like `build.Dockerfile`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@
 #
 # Prerequisite: a working speculos build.
 
-FROM python:3.8-slim
+FROM docker.io/library/python:3.9-slim
 
 ADD . /speculos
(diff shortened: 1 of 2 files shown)


greenknot (approved):

greenknot:
Merged, thanks.

PR #173, opened 2021-06-15
Fix misspellings found using codespell

niooss-ledger:
`codespell` is a tool which looks for common misspellings in source code. It can be used to check Speculos code with:
    
        git ls-files | grep -v '^api/swagger/' | \
            xargs codespell --ignore-words=.codespell-ignore --check-filenames

Fix the misspellings found using it and add an action in GitHub Actions CI to automatically run it in Pull Requests.

Here is an example of failure reported by codespell: https://github.com/niooss-ledger/speculos/runs/2830337818?check_suite_focus=true#step:8:19

### DIFF
--- a/.codespell-ignore
+++ b/.codespell-ignore
@@ -1,0 +1,4 @@
+keypair
+ontop
+tolen
+unstall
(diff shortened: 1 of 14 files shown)


greenknot (approved):

greenknot:
Merged (1211ba763e53f2aeab7a74d0cab63bff18d55cae), thanks!

PR #174, opened 2021-06-15
os_eip2333: remove unnecessary condition

niooss-ledger:
As `path[i]` is an `unsigned int`, it cannot be greater than `0xffffffff`. So the check in function `cx_derive_child_sk` was not
needed, and triggered a warning in LGTM: https://lgtm.com/projects/g/LedgerHQ/speculos/snapshot/7839414ae493f551fbe0850bb249b7617cb46c2d/files/src/bolos/os_eip2333.c?sort=name&dir=ASC&mode=heatmap#L179

Remove this check to silence LGTM.


greenknot:
Merged, thanks.

PR #180, opened 2021-06-16
mcu/apdu: fix forwarding packets without SW

niooss-ledger:
Some applications (like FIDO2/U2F) craft APDU responses with no status word. With makes `mcu/apdu.py` explodes with:

    OverflowError: can't convert negative int to unsigned

Wrap the computed `size` to an unsigned 32-bit value to avoid this.

### DIFF
--- a/mcu/apdu.py
+++ b/mcu/apdu.py
@@ -78,7 +78,7 @@ class ApduClient:
 
         self.logger.info("< {}".format(packet.hex()))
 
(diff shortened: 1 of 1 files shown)


greenknot (approved):

greenknot:
Merged (e9e5356f8aeeb00130f1e585df7b9006ca96a16d), thanks!

PR #191, opened 2021-06-17
Api apdu

kingofpayne:


api/api.py
```diff
@@ -174,6 +176,45 @@ def post(self):
...
+class APDU:
+    def __init__(self):
+        # We want to be notified when APDU response is transmitted from the SE
+        self.condition = threading.Condition()
+
+    def exchange(self, data: bytes) -> bytes:
+        with self.condition:  # Lock for a command/response for one client
+            self.response = None
+            seph.to_app(data)
+            with self.condition:
```
niooss-ledger, line 188: Why is the function locking the same condition twice?

api/api.py
```diff
@@ -174,6 +176,45 @@ def post(self):
...
+                while self.response is None:
+                    self.condition.wait()
+            return bytes(self.response)  # clone before leaving lock
+
+    def seph_apdu_callback(self, data: bytes):
+        """
+        Called by seph when data is transmitted by the SE. That data should
+        be the response to a prior APDU request
+        """
+        self.response = data
```
niooss-ledger, line 198: * Maybe this code should also lock `self.condition` before setting `self.response`
* If `self.response` is not `None`, this means that no one is expecting the response or that a response was already received and will be overwritten. Showing a warning in Speculos console in such a case would help debug unexpected issues. What do you think?

greenknot (changes requested):
Thanks a lot for the PR @kingofpayne!

api/api.py
kingofpayne: This is a mistake, I think first lock should be a separate `threading.Lock` since `wait` releases `condition` until response is available. Would it be ok for you ?

api/api.py
```diff
@@ -174,6 +176,45 @@ def post(self):
...
+class APDU(Resource):
+    def __init__(self):
+        super().__init__()
+        self.parser = reqparse.RequestParser()
+        self.parser.add_argument("data", type=str, required=True)
+
+    def post(self):
+        args = self.parser.parse_args()
+        data = bytes.fromhex(args.get("data"))
+        return {"data": hexlify(apdu.exchange(data)).decode()}
```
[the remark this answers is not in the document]
kingofpayne: Didn't know that! Much better indeed! Fixed.

tests/api/test_api.py
```diff
@@ -138,3 +138,11 @@ def test_screenshot(self, app):
             assert response.status_code == 200
             assert response.headers["Content-Type"] == "image/png"
             assert response.content.startswith(b"\x89PNG")
+
+    def test_apdu(self, app):
+        # Send GET_RANDOM APDU to get 16 bytes of random
+        with requests.post(f"{API_URL}/apdu", json={"data": "e0c0000010"}) \
```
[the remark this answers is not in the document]
kingofpayne: Fixed

api/api.py
niooss-ledger: What is the semantics of `self.condition`? There seems to be 2 possible meanings:
1. a lock on `self.response`, associated with signaling so that `exchange` can wait for `seph_apdu_callback` to receive response data.
2. a lock on the `/apdu` endpoint so that only one client can use it.

If the second meaning was expected, the current code is buggy, because the lock is released when `exchange` waits for the response data. A dedicated lock is then needed, for this second meaning.

Otherwise if the second meaning is not expected (is it possible to send another command before a response was received?), the comment `Lock for a command/response for one client` should be removed.

api/api.py
```diff
@@ -174,6 +176,45 @@ def post(self):
...
+
+
+class APDU(Resource):
+    def __init__(self):
+        super().__init__()
+        self.parser = reqparse.RequestParser()
+        self.parser.add_argument("data", type=str, required=True)
+
+    def post(self):
+        args = self.parser.parse_args()
```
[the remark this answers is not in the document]
kingofpayne: Schema added.

api/swagger.yaml
```diff
@@ -8,6 +8,26 @@ servers:
   - url: "http://127.0.0.1:5000"
 
 paths:
+  /apdu:
+    post:
+      summary: "Transmit APDU query and return device response"
+      parameters:
+      - name: "data"
```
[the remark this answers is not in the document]
kingofpayne: Agreed, changed!

api/api.py
kingofpayne: Both 1 and 2 are expected. `self.condition` should satisfy response waiting, and another `self.lock` should satisfy endpoint locking. I can commit a change if you do agree? (actually I had those two locks at the beginning and thought by mistake I could use only one).

api/api.py
kingofpayne: Fixed

api/api.py
kingofpayne: Fixed. Thank you niooss for helping fixing this part.

api/swagger.yaml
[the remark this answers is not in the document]
greenknot: The parameters are still passed in the query, aren't they?

greenknot (approved):
LGTM :+1: 

Please squash all your commits together and rebase on master before merging.

PR #196, opened 2021-06-28
Fix screenshot race condition

kingofpayne:
Screenshot API requests may return invalid images due to a race condition on the screenshot image update.
This happens frequently especially when recording a GIF.

Here is an example of invalid screenshot during a GIF recording:
speculos


greenknot (approved):
Nice!

PR #205, opened 2021-07-05
mcu: send one text event per line on Nano X

greenknot:
This is already the case on Nano S.

### DIFF
--- a/mcu/nanox_ocr.py
+++ b/mcu/nanox_ocr.py
@@ -1,4 +1,4 @@
-from typing import List, Tuple, Mapping
+from typing import List, Mapping
 from dataclasses import dataclass
 import functools
 import string
@@ -129,16 +129,15 @@ def find_char_from_bitmap(bitmap: BitMap):
                 if all(b == 0 for b in residual_bytes):
                     all_values.append(character_value)
         if all_values:
-            return max([x for x in all_values])
+            char = max([x for x in all_values])
+            if char == "\x80":
+                char = " "
+            return char
 
 
 class NanoXOCR:
     def __init__(self):
-        self.reset()
-
-    def reset(self):
-        self.last_y = 0
-        self.events = [TextEvent("", 0, 0)]
+        self.events = []
 
     def analyze_bitmap(self, data: bytes):
         if data[0] != 0:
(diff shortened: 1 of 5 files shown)


mcu/seproxyhal.py
```diff
@@ -187,8 +187,10 @@ def _send_packet(self, tag, data=b''):
 
     def apply_automation(self, text, x, y):
         if self.automation_server:
-            event = {"text": text.decode("ascii", "ignore"), "x": x, "y": y}
-            self.automation_server.broadcast(event)
+            lines = text.decode("ascii", "ignore").split("\n")
+            for line in lines:
+                event = {"text": line, "x": x, "y": y}
```
niooss-ledger, line 192: I am not familiar with this code, but shouldn't `y` be updated for each line?

greenknot: Thanks for that feedback. Indeed, it's wrong. I pushed force a new commit which is a bit more complex but handles correctly that.

mcu/nanox_ocr.py
```diff
@@ -151,19 +151,20 @@ def analyze_bitmap(self, data: bytes):
...
 
-        self.last_x, self.last_y = x, y
+        self.last_y = y
 
-    def get_text(self, reset=True) -> Tuple[bytes, Tuple[int, int]]:
-        text, x, y = self.text, self.x, self.y
-        if reset:
-            self.reset()
-        return text, (x, y)
+    def get_events(self) -> Tuple[bytes, Tuple[int, int]]:
```
niooss-ledger, line 167: The return type must be updated

mcu/nanox_ocr.py
```diff
@@ -152,18 +151,15 @@ def analyze_bitmap(self, data: bytes):
...
-        if reset:
-            self.reset()
-        return text, (x, y)
+            # create a new TextEvent if there are no events yet or if there is a new line
+            if not self.events or y > self.events[-1].y:
+                event = TextEvent("", x, y)
+                self.events.append(event)
+            else:
+                event = self.events[-1]
+            event.text += char
```
niooss-ledger, line 160: Or even simpler:
```python
            if not self.events or y > self.events[-1].y:
                self.events.append(TextEvent(char, x, y))
            else:
                self.events[-1].char += char
```

... which becomes even more readable by inversing the condition:

```python
            if self.events and y <= self.events[-1].y:
                self.events[-1].char += char
            else:
                self.events.append(TextEvent(char, x, y))
```

niooss-ledger (approved):

PR #206, opened 2021-07-05
Save caller try_context to avoid anarchic longjmps

Saltari:
In the event of a `os_lib_call`, the caller `try_context` can be erased by the `try_context` of the callee as its main is likely to be in a `TRY/CATCH` block. After returning to the caller with a `os_lib_end` if the call triggers an exception of some sorts (`THROW`), the `longjump` performed to branch to the `try_context` stack pointer uses a wrong address (the one of the stack pointer's callee's `try_context`) This commit aims at saving/restoring the caller's context after a call to `os_lib_call` to prevent the issue

### DIFF
--- a/src/bolos/os.c
+++ b/src/bolos/os.c
@@ -20,6 +20,7 @@
 struct libcall_s {
   struct app_s *app;
   struct sigcontext sigcontext;
+  try_context_t *trycontext;
 };
 
 static try_context_t *try_context;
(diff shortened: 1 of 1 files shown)


greenknot (approved):
Thanks a lot @Saltari! It LGTM :+1: 

I added @pscott as an additional reviewer since he's an heavy user of the `os_lib_call` mechanism and could tell if it works for his usecases.

pscott:
I've tested this yesterday and today with a couple of plugins and it solved every issues I had. I will test again today with a couple of additional testing inputs, but I'm fairly certain this fix is correct :D

pscott:
Can confirm this solves pretty much all the plugin test cases I've tried, without inducing any weird behavior.

:+1: merge! :D

greenknot:
Many thanks to both of you!

PR #209, opened 2021-07-07
api: on /apdu, send HTTP headers before having the exchange response

greenknot:
It avoids HTTP client being blocked. Eg.

```
  with requests.post(f"{API_URL}/apdu", json={"data": data}) as r:
    print("do stuff even in the response isn't available yet")
```

### DIFF
--- a/api/api.py
+++ b/api/api.py
@@ -8,6 +8,7 @@ from PIL import Image
 import pkg_resources
 import threading
 import time
+from typing import Generator
 
 import mcu.automation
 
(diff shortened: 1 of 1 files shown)


api/api.py
```diff
@@ -183,14 +183,17 @@ def __init__(self):
         self.response_condition = threading.Condition()
 
     def exchange(self, data: bytes) -> bytes:
```
niooss-ledger, line 185: The return type needs to be updated (with `Generator[bytes, None, None]`? cf. https://docs.python.org/3/library/typing.html#typing.Generator)

kingofpayne:
Tested without any issue. Just update the return type and it's ok for me.

kingofpayne (approved):

PR #212, opened 2021-07-08
Move Python files to speculos/ and add setup.py

niooss-ledger:
This Pull Request completes the work initiated in https://github.com/LedgerHQ/speculos/pull/120 to be able to use `pip` to install Speculos. With this, using `pip install .` in the root directory of this project builds the launcher (using `cmake`) and install a `speculos` Python package, as well as a `speculos` program which lauches Speculos.

This PR:

* Modifies the `cmake` configuration to copy the built launcher and VNC server to `speculos/resources`. This enables using `cmake` with a temporary build directory which is not `build/`. Moreover this enables to package the produced binaries without embedding all the files generated by `cmake` and the C compilation process.
* Moves all Python files in a new `speculos/` directory. This is needed in order to easily install a Python package named `speculos` (otherwise `api` and `mcu` would be seen as top-level packages). Doing this, some imports were modified in order to use relative imports.
* Adds a `setup.py` (with `MANIFEST.in` to install binary files) to package Speculos.

With this, both `python3 setup.py install` and `pip install .` work, to install Speculos in a Python installation.

PR #213, opened 2021-07-08
speculos/api: plug the API to the notifiers mechanism

niooss-ledger:
*This PR depends on https://github.com/LedgerHQ/speculos/pull/212 as this works was done after it.*

When the API server fails to start, for example when there is already a server bound to TCP port 5000, Speculos continued to run. This is because the Flask web server was launched in a separate thread and propagating a failure is not straightforward.

Introduce a new `socketpair` to signal Speculos' main event loop that the API server stopped and that Speculos should exits.

### DIFF
--- a/speculos/api/api.py
+++ b/speculos/api/api.py
@@ -6,11 +6,14 @@ import io
 import os.path
 from PIL import Image
 import pkg_resources
+import socket
 import threading
 import time
 from typing import Generator
 
 from ..mcu import automation as mcu_automation
+from ..mcu.readerror import ReadError
+
 
 static_folder = pkg_resources.resource_filename(__name__, "/swagger")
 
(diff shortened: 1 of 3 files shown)


greenknot (changes requested):
Cool, thanks!

If `bind` fails, speculos sometimes exits without any error message. Maybe something like this could be helpful, what do you think?

```diff
diff --git a/speculos/api/api.py b/speculos/api/api.py
index 7b8a139..9df614c 100644
--- a/speculos/api/api.py
+++ b/speculos/api/api.py
@@ -1,5 +1,6 @@
 from flask import Flask, request, stream_with_context, Response
 from flask_restful import inputs, reqparse, Resource, Api
+import errno
 import json
 import jsonschema
 import io
@@ -68,6 +69,11 @@ class ApiRunner:
         try:
             # threaded must be set to allow serving requests along events streaming
             app.run(host="0.0.0.0", port=self.api_port, threaded=True, use_reloader=False)
+        except OSError as e:
+            if e.errno == errno.EADDRINUSE:
+                app.logger.error(f"address already in used (0.0.0.0:{self.api_port})")
+            else:
+                raise
         finally:
             self._notify_exit.close()
 
```

niooss-ledger:
Thanks for the review. In fact the issue is more generic: if Flask raises an exception, it can be eaten by the main process exiting too fast.

I added https://github.com/LedgerHQ/speculos/pull/213/commits/a8f79c03bbfd58ad97a1a857b0b302ecffd2f1e3 to fix this issue.

greenknot (approved):

PR #216, opened 2021-07-09
Minor bug fix in APDU API printing invalid warning

kingofpayne:


speculos/api/api.py
```diff
@@ -202,9 +202,6 @@ def seph_apdu_callback(self, data: bytes):
         be the response to a prior APDU request
         """
         with self.response_condition:
-            if self.response_condition is not None:
```
kingofpayne, line 205: This line was also wrong by the way.

kingofpayne:
Tests in CI are using the legacy way to transmit APDUs, so the check would always print warnings, which is not good.

PR #220, opened 2021-07-13
Publish Speculos on TestPyPI in the CI

niooss-ledger:
Use commands from https://packaging.python.org/tutorials/packaging-projects/ to build the Python package and push it to TestPyPI at each commit on `master`.
    
This requires the C code of Speculos to be part of the source tarball ("sdist"). Modify `MANIFEST.in` accordingly.
As this also requires that `setup.py` no longer builds the C tests, this PR depends on https://github.com/LedgerHQ/speculos/pull/219.


greenknot (approved):
Nice!

greenknot (approved):

Issue #222, opened 2021-07-16
Display the screen on the web server


greenknot:
On MacOS, VNC is the only mechanism to display the screen of the emulated device.

It would be great to add a web page (`index.html`?) to the web server which make use of the REST API to:

- display the screen
- interact with the device (buttons, maybe APDUs?)

The script `tools/gif-recorder.py` shows how the API can be used to get notified when the screen changes.

greenknot:
Done (#223).

PR #223, opened 2021-07-16
Add a basic display to the web interface

greenknot:
See #222.

The web interface is available at http://127.0.0.1:5000/display.html and here is a screenshot:

Screenshot from 2021-07-16 12-18-44

The UI is terrible. Need someone with web skillz :)


greenknot:
Quick update to make it less atrocious:
 
screenshot-web

The web interface is available at `/` and swagger at `/swagger`. The `left`, `right` and `up` keys can be used instead of the buttons.

I think it's already quite useful, but any enhancement is welcome.

greenknot:
On a side note, this doesn't work well with the Blue device (because the screen is too large, and finger actions aren't implemented).

TrAyZeN:
> Quick update to make it less atrocious:
>  
> screenshot-web
> 
> The web interface is available at `/` and swagger at `/swagger`. The `left`, `right` and `up` keys can be used instead of the buttons.
> 
> I think it's already quite useful, but any enhancement is welcome.

Hi, it's great. As an enhancement, I think it would be better to put the `both` button in the middle.

niooss-ledger:
> Hi, it's great. As an enhancement, I think it would be better to put the both button in the middle.

Done, as part of https://github.com/LedgerHQ/speculos/pull/223/commits/507dc73949e576305e75611b6080aea2d6a53f5f ("api: add alt and title texts").

I directly modified the Pull Request with some improvements, and with them, this looks good to me :+1:

niooss-ledger (approved):

PR #225, opened 2021-07-21
Introduce version 0.1

niooss-ledger:
This prepares the publication of Speculos package on PyPI and TestPyPI. The idea is to tag the merge commit of this PR with `0.1`. Then the version of each merge commit of the master branch can be computed with:

```sh
LAST_TAG="$(git describe --tags --abbrev=0)"
COUNT="$(git rev-list --count "${LAST_TAG}..HEAD")"
VERSION="${LAST_TAG}.${COUNT}"
```

(For example the merge commit 10 commits after tag `0.1` would be version `0.1.10`.

Introduce `tools/update_setup_version_from_git.sh` to do this in `setup.py`. This file may then be called from GitHub Action to update `setup.py` before creating the Python package (I will update https://github.com/LedgerHQ/speculos/pull/220 accordingly once this PR is discussed/merged).

### DIFF
--- a/tools/update_setup_version_from_git.sh
+++ b/tools/update_setup_version_from_git.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+# Update the version field of Speculos python package, in setup.py.
+#
+# This script is intended to be used before creating a Python package, in order
+# to create a source tarball ("sdist package") with a version which has been
+# defined using the number of git commits since the last tag, without requiring
+# to have the git repository in the package.
+#
+# Git tags are expected to be "major.minor", such as "0.1".
+
+set -eu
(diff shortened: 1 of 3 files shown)


tools/update_setup_version_from_git.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
```
greenknot, line 2: It might be helpful to add a few comments here to tell what's the purpose of this script and what's the expected output.

niooss-ledger: I added a comment.

greenknot (approved):

PR #229, opened 2021-07-21
Add an API Python client

greenknot:


greenknot:
The main goal of this PR is to introduce a Python file which can be imported by other projects (`from speculos import client`) to use the API, for example in test scenarios.

Speculos tests (eg. `tests/apps/test_btc.py`) were updated to make use of this file.

speculos/client.py
```diff
@@ -0,0 +1,181 @@
...
+            except ConnectionRefusedError:
+                time.sleep(0.1)
+
+        if not connected:
+            raise f"Failed to connect to the speculos instance on port {self.port}"
+
+        s.close()
+
+    def start(self):
+        cmd = ["python3", "-m", "speculos.main"] + self.args + [self.app]
```
niooss-ledger, line 138: `"python3"` should be replaced with `sys.executable` (https://docs.python.org/2.7/library/sys.html#sys.executable)

speculos/client.py
```diff
@@ -0,0 +1,181 @@
...
+            assert response.headers["Content-Type"] == "image/png"
+            return response.content
+
+    def _apdu_exchange(self, data: bytes) -> bytes:
+        with self.session.post(f"{self.api_url}/apdu", json={"data": data.hex()}) as response:
+            apdu_response = ApduResponse(response)
+            return apdu_response.receive()
+
+    def _apdu_exchange_async(self, data: bytes) -> requests.Response:
+        return self.session.post(f"{self.api_url}/apdu", json={"data": data.hex()}, stream=True)
```
niooss-ledger, line 100: "async" is misleading, in the context of Python: `async` has a specific meaning, related to AsyncIO (https://docs.python.org/3/library/asyncio.html).

I suggests other names: `apdu_exchange_nopoll`, `apdu_exchange_nowait`, `apdu_exchange_request`

speculos/client.py
```diff
@@ -0,0 +1,181 @@
...
+        if status != 0x9000:
+            raise ApduException(status)
+        return data
+
+
+class Api:
+    def __init__(self, api_url: str) -> None:
+        self.api_url = api_url
+        self.timeout = 2000
+        self.session = requests.session()
```
niooss-ledger, line 54: `requests.session()` is deprecated (https://github.com/psf/requests/commit/1c13c5e6f35cf05c4de758714bc99919e9bbbb43) and `requests.Session()` should be used instead

speculos/client.py
```diff
@@ -0,0 +1,181 @@
...
+            event = self.get_next_event()
+            if text in event["text"]:
+                break
+        return event
+
+    def press_and_release(self, button: str):
+        assert button in ["left", "right", "both"]
+        data = {"action": "press-and-release"}
+        with self.session.post(f"{self.api_url}/button/{button}", json=data) as response:
+            assert response.status_code == 200
```
niooss-ledger, line 81: Using `assert` to check a condition which can happen (for example when Speculos's API server is down...) is not developer-friendly. Moreover if this happens, developers have no way of knowing what went wrong from their Python code.

Could this be replaced with raising an exception (which would mean "an HTTP error occurred: ...") instead?

speculos/client.py
```diff
@@ -0,0 +1,181 @@
...
+    return data[:-2], status
+
+
+def screenshot_equal(path1: str, path2: str) -> bool:
+    """Compare two images and return True if they are equal."""
+
+    with Image.open(path1) as img1:
+        with Image.open(path2) as img2:
+            diff_img = ImageChops.difference(img1, img2)
+    return diff_img.getbbox() is None
```
niooss-ledger, line 27: [I am not sure about my comment, feel free to ignore it]

Using `getbbox` seems to be overkill, just to check whether an image contains only black pixels. Another possibility would be to use `getcolors` (https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.Image.getcolors):

```python
colors = diff_img.getcolors(maxcolors=1)
return colors is not None and len(colors) == 1 and colors[0][1] == (0, 0, 0)
```
What do you think?

greenknot: Discussed offline and the current code seems to be more maintainable in the long term.

speculos/client.py
```diff
@@ -0,0 +1,193 @@
...
+        self.process: Optional[subprocess.Popen] = None
+
+        if "--display" not in self.args:
+            self.args += ["--display", "headless"]
+
+        if "--api-port" not in self.args:
+            self.port = 5000
+        else:
+            n = self.args.index("--api-port")
+            self.port = int(self.args[n])
```
niooss-ledger, line 131: This function has two bugs:

1. Using `args=[]` as parameter has the unexpected side effet of declaring a static list associated with the function. So every time `__init__` is called without specifying `args`, parameters are *added* to `args`. Moreover if a caller calls the constructor with code such as:
```python
my_args = ["--apt-port", "5000"]
instance1 = SpeculosInstance("app", my_args)
my_args[1] = "5001"
instance2 = SpeculosInstance("app", my_args)
```
... in fact `my_args[1] = "5001"` will modify `instance1.args` unexpectedly. I suggest modifying the code to:

```python
from typing import Iterable
# ...
class SpeculosInstance:
    def __init__(self, app: str, args: Iterable[str] = ()) -> None:
        self.app = app
        self.args = list(args)
# ...
```

2. `self.port = int(self.args[n])` calls `int` on `"--api-port"` instead of the API port value.  `self.port = int(self.args[n + 1])` should be used instead.

tests/apps/conftest.py
```diff
@@ -36,55 +36,92 @@ def list_apps_to_test(app_dir) -> List[AppInfo]:
...
-    _app = APDUClient(request.param)
-    yield _app
-    _app.stop()
+def app(request, client):
+    return app_info_from_path(client.app)
 
 
-def pytest_generate_tests(metafunc):
+def get_apps(name):
     # retrieve the list of apps in the ../apps directory
```
niooss-ledger, line 60: This should be a function docstring

tests/apps/conftest.py
```diff
@@ -36,55 +36,92 @@ def list_apps_to_test(app_dir) -> List[AppInfo]:
...
+
+def default_btc_app():
+    app_dir = os.path.join(SCRIPT_DIR, os.pardir, os.pardir, "apps")
+    filepath = os.path.realpath(os.path.join(app_dir, "btc.elf"))
+    apps = get_apps("btc")
+    return [app for app in apps if os.path.realpath(app.filepath) == filepath]
+
+
+def idfn(app):
+    return os.path.basename(app.filepath)
```
niooss-ledger, line 75: What is `idfn`? A function docstring pointing to pytest's documentation or anything useful to understand what it is about (Is `basename` an implementation detail or does this function really needs to return the name of a file? Is the output of the function displayed to the user in some circumstances, which requires its return value to be quite meaningful?)

tests/apps/conftest.py
```diff
@@ -36,55 +36,92 @@ def list_apps_to_test(app_dir) -> List[AppInfo]:
...
+
+
+def client_instance(app, additional_args=[]):
+    args = ["--model", app.model, "--sdk", app.sdk] + additional_args
```
niooss-ledger, line 79: Using `...=[]` is dangerous in a function declaration. Please use `=()` or `=None`, whatever is the more appropriate.

niooss-ledger (changes requested):
I finished my review. It looks good to me, but some (minor/easy) issues need to be fixed before merging this. Thanks for this feature!

tests/apps/conftest.py
greenknot: From the [documentation](https://docs.pytest.org/en/6.2.x/example/parametrize.html#different-options-for-test-ids) about the pytest fixtures `ids` parameter:

> pytest will build a string that is the test ID for each set of values in a parametrized test. These IDs can be used with -k to select specific cases to run, and they will also identify the specific case when one is failing. Running pytest with --collect-only will show the generated IDs.

```console
$ python3 -m pytest tests/apps/ -v
...            
tests/apps/test_btc.py::test_btc_get_version[nanos#btc#1.5#00000000.elf] PASSED                        [  4%]
tests/apps/test_btc.py::test_btc_get_public_key_with_user_approval[nanos#btc#1.5#00000000.elf] SKIPPED [  8%]
tests/apps/test_btc.py::test_btc_automation[nanos#btc#1.5#00000000.elf] SKIPPED                        [ 12%]
tests/apps/test_btc.py::test_btc_get_version[nanos#btc#1.6#6bd0a5f8.elf] PASSED                        [ 16%]
tests/apps/test_btc.py::test_btc_get_public_key_with_user_approval[nanos#btc#1.6#6bd0a5f8.elf] PASSED  [ 20%]
```

This is much useful than (previously):

```console
$ python3 -m pytest tests/apps/ -v
...
tests/apps/test_btc.py::test_btc_get_version[client_btc0] PASSED                        [  4%]
tests/apps/test_btc.py::test_btc_get_public_key_with_user_approval[client_btc0] SKIPPED [  8%]
tests/apps/test_btc.py::test_btc_automation[client_btc0] SKIPPED                        [ 12%]
tests/apps/test_btc.py::test_btc_get_version[client_btc1] PASSED                        [ 16%]
tests/apps/test_btc.py::test_btc_get_public_key_with_user_approval[client_btc1] PASSED  [ 20%]
tests/apps/test_btc.py::test_btc_automation[client_btc1] PASSED                         [ 24%]
```

speculos/client.py
greenknot: Nice catch!

niooss-ledger (approved):

PR #232, opened 2021-07-23
Fix event stream

greenknot:
Server-Sent Events (SSE) enables a client to receive automatic updates from a
server via an HTTP connection. It is now part of HTML5 since a few years and
supported by all browsers but IE:

- https://www.w3.org/TR/eventsource/
- https://html.spec.whatwg.org/multipage/server-sent-events.html

If fixes a bug where some events were received twice by the web page on Chrome
and Firefox.

### DIFF
--- a/tests/api/test_api.py
+++ b/tests/api/test_api.py
@@ -2,6 +2,7 @@ import json
 import os
 import pkg_resources
 import pytest
+import re
 import requests
 from collections import namedtuple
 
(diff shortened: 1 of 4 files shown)


tests/api/test_api.py
```diff
@@ -113,18 +113,28 @@ def test_events(self, app):
...
         """
 
+        def get_next_event(stream):
+            data = b""
+            while True:
+                line = stream.raw.readline()
+                if line == b"\n":
+                    break
+                assert line.startswith(b"data: ")
+                data += line[6:-1]
```
niooss-ledger, line 123: This `while` loop seems strange, as it merges several lines starting with `data: ` until an empty line is read, and then callers call `json.loads(data)` on the result. If this is the expected behavior, this should be documented as a comment.

Moreover if all callers are expected to call `json.loads`, it would make sense to move such calls inside `get_next_event`.

niooss-ledger (approved):
With this change, it is now possible to see the EventStream in the Network tab of Chromium Developer Tools. Nice :)

Issue #230, opened 2021-07-22
Versioning scheme for Speculos package


niooss-ledger:
Hello,
As Speculos is now available on [PyPI](https://pypi.org/project/speculos/) and [TestPyPI](https://test.pypi.org/project/speculos/), some decisions have to be made on how versions are defined in Speculos.

For TestPyPI, every commit on `master` is versioned as `maj.min.count` where `maj.min` is the last tag and `count` the number of commits since the last tag. This was implemented in https://github.com/LedgerHQ/speculos/pull/225 and https://github.com/LedgerHQ/speculos/pull/220. This work is/will be integrated with the CI, and is simple enough to understand.

However this does not fit the "semantic versioning" principles. For example when a breaking change is introduced in version 1.0.42 (for example "let's remove feature A for Speculos"), this would automatically create a version 1.0.43 when this change is merged, instead of version 1.1.0 or 2.0.0. (Then a git tag could be added to Speculos to create a proper release).

I see mainly two ways of solving this:

* Only create releases (on PyPI) when a git tag is created. This would prevent a 1.0.43 from being created (and the CI would create such a version on TestPyPI for testing purposes). This would require the project maintainers to regularly tag a new release, which would introduce "more overhead".
* Automatically release versions of `master` branch on PyPI like on TestPyPI and create a way to stop releasing new versions on PyPI when a breaking change occurs (for example the version in `setup.py` could be bumped to 2.0.0 and this it does not match the last git tag, the release is not done). This would introduce more complexity in the version handling process, which could become more fragile.

There could be other ways to fix this. Let's discuss this in this issue!

greenknot:
Thanks @niooss-ledger for tackling this topic.

Another idea which doesn't involve tags: just use the version from `setup.py`. I don't know if it's considered good practice, but tags are of no use currently and introduce some complexity in the release process.

The following command-line:

```console
$ git rev-list --count "$(git blame -L$(git grep -h --line-number version= setup.py | cut -d : -f 1),+1 setup.py | cut -d ' ' -f 1)..HEAD"
```

1. Retrieves the line number of `setup.py` which contains `version=` thanks to `git grep`
2. Passes it to `git blame` to get the revision when this line was modified last
3. Eventually get the number of commits since this last revision

niooss-ledger:
[...]

PR #234, opened 2021-07-29
tools/update_setup_version_from_git: no longer rely on git tag to modify the version

niooss-ledger:
As Speculos git repository does not use tags currently, the reference of the version of the Python package is `setup.py`.

Change the way versions are computed so that it is `x.y.z` where the `setup.py` in git contains `x.y.0` and `z` is computed by counting the number of commits since the last change of the version line.

Suggested-in: https://github.com/LedgerHQ/speculos/issues/230

### DIFF
--- a/tools/update_setup_version_from_git.sh
+++ b/tools/update_setup_version_from_git.sh
@@ -3,16 +3,21 @@
 #
 # This script is intended to be used before creating a Python package, in order
 # to create a source tarball ("sdist package") with a version which has been
-# defined using the number of git commits since the last tag, without requiring
-# to have the git repository in the package.
+# defined using the number of git commits since the last time the package
+# version was modified in setup.py, without requiring to have the git repository
+# in the package.
 #
 # Git tags are expected to be "major.minor", such as "0.1".
 
 set -eu
 cd "$(dirname -- "${BASH_SOURCE[0]}")/.."
 
-LAST_TAG="$(git describe --tags --abbrev=0)"
-COUNT="$(git rev-list --count "${LAST_TAG}..HEAD")"
-VERSION="${LAST_TAG}.${COUNT}"
+VERSION_LINE="$(git grep -h --line-number '^    version=".*",$' setup.py | cut -d : -f 1)"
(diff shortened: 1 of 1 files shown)


tools/update_setup_version_from_git.sh
```diff
@@ -3,16 +3,20 @@
 #
 # This script is intended to be used before creating a Python package, in order
 # to create a source tarball ("sdist package") with a version which has been
-# defined using the number of git commits since the last tag, without requiring
-# to have the git repository in the package.
+# defined using the number of git commits since the last time setup.py was
```
greenknot, line 6: "since the last time the package version was modified in setup.py" maybe?

tools/update_setup_version_from_git.sh
```diff
@@ -3,16 +3,20 @@
...
 
-LAST_TAG="$(git describe --tags --abbrev=0)"
-COUNT="$(git rev-list --count "${LAST_TAG}..HEAD")"
-VERSION="${LAST_TAG}.${COUNT}"
+VERSION_LINE="$(git grep -h --line-number '^    version=".*",$' setup.py | cut -d : -f 1)"
+LAST_CHANGE="$(git blame --abbrev=40 -L"$VERSION_LINE",+1 HEAD setup.py | cut -d ' ' -f 1)"
+COUNT="$(git rev-list --count "${LAST_CHANGE}..HEAD")"
 
-sed 's/^\( *version=\)"[^"]*"/\1"'"${VERSION}"'"/' -i setup.py
+sed 's/^\(    version=\)"\([0-9]\+\.[0-9]\+\)\.[0-9]\+"/\1"\2.'"$((COUNT - 1))"'"/' -i setup.py
```
greenknot, line 18: `COUNT-1` is invalid if `COUNT` is zero. It can happen if the last commit changed `setup.py`.

niooss-ledger: Indeed, I misunderstood why this new algorithm differed by 1 with the current one, but it is been the tag `0.1` was applied on commit d978cadebd2ba006b728dafedce550182b6be3c9, which is the merge commit of the Pull Request which modified the version in `setup.py` in 04b6101100a24e4211a1d16803a3444a6d200402. So `COUNT` should be used as-is. I fixed my code.

greenknot (approved):

PR #235, opened 2021-07-29
Added 'data' argument to ApduException; made args mandatory

bigspider:
The ApduException class is missing information about the returned APDU data. This PR adds it to the constructor in order to enable handling the error in a more precise way (in my case, I return a special SW that is not an error, but rather a request for the client side; therefore, it is crucial to have access to the apdu data).


niooss-ledger (approved):

greenknot (approved):
LGTM, thanks @bigspider!

Issue #236, opened 2021-07-30
API client status word checking improvements


kingofpayne:
Current `apdu_exchange` and `apdu_exchange_nowait` methods of `SpeculosClient` strips off and check automatically the status word from the device response.

https://github.com/LedgerHQ/speculos/blob/96875a8b66237dd280ada8545fbd273e56701dff/speculos/client.py#L38-L43

In particular, it only checks that the status word is `0x9000`, and there are other status words which should not be considered as errors, such as: `0x61XX`, `0x9100`, `0x9200` (as described in https://www.eftlab.com/knowledge-base/complete-list-of-apdu-responses/).

In a corner case example, a test may want to check that the returned status word is `0x61XX` (and verify `XX` has the correct value) and consider `0x9000` as invalid, which I is not possible currently. This status word also conveys information that must be retrieved by the caller.

Delegating the status word check to the client is very convenient in most cases, but this shall not be mandatory.

Side note: Tests using speculos may want to verify that error status words are correctly returned in defined cases. This can be done currently by catching the exception and inspecting it.

PR #237, opened 2021-07-30
Proposal for #236

kingofpayne:


speculos/client.py
```diff
@@ -33,14 +33,32 @@ def check_status_code(response: requests.Response, url: str) -> None:
...
+            self.__response = None
+
+    def data(self, expected_sw: int=0x9000) -> bytes:
+        """
+        :return: Received data (status word stripped)
+        :param expected_sw: Expected status word value.
+        :raises ApduException: If received status word is different from `expected_sw` parameter.
+        """
+        self.__receive()
+        if self.__status != 0x9000:
```
niooss-ledger, line 52: This should be `if self.__status != expected_sw:`, if I understand correctly the aim of this Pull Request.

speculos/client.py
```diff
@@ -33,14 +33,32 @@ def check_status_code(response: requests.Response, url: str) -> None:
...
-        if status != 0x9000:
-            raise ApduException(status, data)
-        return data
+        self.__response = response
+
+    def __receive(self):
+        """ Get the response if not yet received and store it in `__data` and `__status` attributes """
+        if self.__response is not None:
+            check_status_code(self.__response, "/apdu")
+            self.__data, self.__status = split_apdu(bytes.fromhex(self.__response.json()["data"]))
```
niooss-ledger, line 42: [Code style] I personally prefer when the function calls are not to much nested. For example I prefer code such as:
```python
json_resp = self.__response.json()
apdu_message = bytes.fromhex(json_resp["data"]))
self.__data, self.__status = split_apdu(apdu_message)
```
... because by doing so, when one of the call fails and raises an exception, the exception call trace indicates precisely which function failed.
Anyway this is my personal opinion, and feel free to follow such advice or keep your code as-is.

speculos/client.py
kingofpayne: Yes, fixed.

speculos/client.py
kingofpayne: We can integrate that change in the PR as well.

kingofpayne:
There's still an issue with this change: status word checking is not performed if `data` method of `ApduResponse` is not called.

PR #244, opened 2021-08-11
Fix #243: Use of Python "with" semantic in API client to guarantee speculos instance termination

### LINKED ISSUE #243: speculos client may not close speculos instance properly
Current speculos client library requires the user to explicitly call the method `stop` to kill running speculos instance. This is a problem for two main reasons:
- the library user may simply forget to do this call,
- an exception may terminate the program without calling the `stop` method.

To circumvent this, a hack has been added
https://github.com/LedgerHQ/speculos/blob/master/speculos/client.py#L188-L189

However, there is no guarantee on the order of cleanup performed by Python when the program terminates, and this call may fail if some used variables (such as the `logger` global variable) are destroyed before `__del__` is called. Here is an example of crash log:

    Exception ignored in: <function SpeculosInstance.__del__ at 0x7fda292da0d0>
    Traceback (most recent call last):
      File "/usr/local/lib/python3.8/dist-packages/speculos/client.py", line 195, in __del__
      File "/usr/local/lib/python3.8/dist-packages/speculos/client.py", line 174, in stop
      File "/usr/lib/python3.8/logging/__init__.py", line 1731, in __repr__
    TypeError: 'NoneType' object is not callable

Maybe the right solution would be to use the `with` semantic of Python to start speculos on `__enter__` and stop it on `__exit__`. This would require modifying slightly the `SpeculosClient` class to support this.

kingofpayne:


speculos/client.py
```diff
@@ -185,15 +185,19 @@ def stop(self) -> None:
...
         SpeculosInstance.__init__(self, app, args)
+        self.__api_url = api_url
+
+    def __enter__(self) -> "SpeculosClient":
         SpeculosInstance.start(self)
-        Api.__init__(self, api_url)
+        Api.__init__(self, self.__api_url)
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb) -> None:
```
greenknot, line 199: Please add type annotation for all arguments.

FYI, it was found with:

```console
$ mypy --strict speculos/client.py
speculos/client.py:199: error: Function is missing a type annotation for one or more arguments
```

unfortunately `mypy` warnings should be fixed across the entire code base before being integrated to the CI.

greenknot (changes requested):
Thanks @kingofpayne. I'm not familiar with https://docs.python.org/3/library/contextlib.html, however it LGTM.

kingofpayne:
Updated `client.py` to start and stop correctly both speculos and event streaming.
It is also now possible to call manually `start` and `stop` on `SpeculosInstance` if the user is not willing to use `with` statement.

Also added typing for `SpeculosInstance.__exit__`.

greenknot (approved):

PR #246, opened 2021-09-08
translate relative paths from windows to linux

yhql:
This partially fixes the rather specific problem of invoking 

```
wsl -e speculos \windows\style\path.elf
```

which would not work otherwise.


greenknot (changes requested):
It works but there are probably better fixes :)

What about using https://docs.python.org/3/library/pathlib.html#basic-use ?

PR #254, opened 2021-11-09
Add support for Brainpool curves

jibeee:
- Define domain parameters for all supported Brainpool curves.
- Add tests for ECDH with Brainpool curves.
- Fix bug in ECDH with CX_ECDH_POINT, which handled 256-bit curves only.
- Fix bug in sys_cx_ecfp_generate_pair2 that was only handling 256-bit curves.


srasoamiaramanana-ledger (approved):

greenknot (approved):
Thanks!

PR #255, opened 2021-11-15
Fix os_perso_seed_cookie with SDK 1.6 & SDK 2.0

dmorais-ledger:


sdk/bolos_syscalls_1.6.h
```diff
@@ -9,8 +9,6 @@
 #define SYSCALL_nvm_write_ID_OUT  0x900003bcUL
 #define SYSCALL_cx_rng_u8_ID_IN  0x600004c0UL
 #define SYSCALL_cx_rng_u8_ID_OUT  0x90000425UL
-#define SYSCALL_cx_rng_u32_ID_IN  0x600089ecUL
```
greenknot, line 12: Is it related to this PR?

dmorais-ledger: As the problem was due to a bad Sycall ID, I checked all those listed in this file and found that SYSCALL_cx_rng_u32_ID_IN was duplicated.

PR #257, opened 2021-11-19
Zeroize app ram in latest FW versions

Saltari:
In latest FW versions, the application RAM (.bss + stack) is zeroized at startup

### DIFF
--- a/src/launcher.c
+++ b/src/launcher.c
@@ -258,9 +258,22 @@ static void *load_app(char *name)
       warn("mmap data");
       goto error;
     }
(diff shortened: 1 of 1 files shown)


greenknot (changes requested):
Thanks @Saltari.

I would add a `default` case for the sake of clarity and to prevent compiler warnings:

```
/* In newer FW versions the app RAM (.bss and stack) is zeroized.
 * This is done by mmap thanks to the MAP_ANONYMOUS flag. */
default:
  break;
```

greenknot (approved):

PR #261, opened 2021-12-01
client: fix error when closing stream

greenknot:
```
  Traceback (most recent call last):
    File "./demo.py", line 125, in <module>
      client.stop()
    File "/speculos/speculos/client.py", line 224, in __exit__
      self.stop()
    File "/speculos/speculos/client.py", line 213, in stop
      self.close_stream()
    File "/speculos/speculos/client.py", line 76, in close_stream
      self.stream.close()
  AttributeError: 'NoneType' object has no attribute 'close'
```

### DIFF
--- a/speculos/client.py
+++ b/speculos/client.py
@@ -73,7 +73,8 @@ class Api:
         check_status_code(self.stream, "/events")
 
     def close_stream(self) -> None:
-        self.stream.close()
+        if self.stream:
+            self.stream.close()
         self.stream = None
 
     def get_next_event(self) -> dict:
--- a/tests/unit/test_client_Api.py
+++ b/tests/unit/test_client_Api.py
@@ -1,0 +1,14 @@
+from unittest import TestCase
+
+from speculos.client import Api
+
+
+class TestApi(TestCase):
+
+    def setUp(self):
+        self.api_url = 'some random url'
+        self.api = Api(self.api_url)
+
+    def test_close_stream_None_should_not_raise(self):
+        self.assertIsNone(self.api.stream)
+        self.api.close_stream()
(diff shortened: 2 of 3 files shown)


greenknot:
Actually, this can't happen unless there is a bug in the code making use of the client. Closing...

lpascal-ledger:
Debatable, this could be considered a legitimate bug in Speculos.

Successive call to `SpeculosClient:stop` would lead to this, including implicit call such as in:
```
with SpeculosClient() as client: 
    client.stop()
```
It could be argued this syntax should not raise.

lpascal-ledger:
In any case, a test should enforce the expected behavior.

lpascal-ledger:
Reopening as the current behavior is unexplicit:
* `Api:open_stream` and `Api:close_stream` are public methods, pulled up the user-oriented `SpeculosClient` class through inheritance and indirectly through public methods `SpeculosClient:start` and `SpeculosClient:stop`. So currently the `stream` lifespan is explicitly manageable by an end-user.
* The context manager on `SpeculosClient` is but a convenient syntactic sugar to enforce the stream to close implicitly.

Usage of both or successive `close` leads to this bug.

I think of 3 responses to that:
* The `with` syntax is the one and only. `start`, `stop`, `open_stream` and `close_steam` should be 'protected' and users only uses them at their own risk.
* Current behavior is OK, users should know when they can close their steam and could cope with exceptions. In this case the exception may benefit from being a bit more explicit (catching the `AttributeError` and raising a `StreamNotInitialized` for instance, or symetrically to `Api:open_steam`, `assert self.stream is not None`).
* All these methods are legit public methods and users should not be bothered if they close the steam twice. This is the choice of this PR.

greenknot:
Thanks a lot @lpascal-ledger for the analysis and the tests.

> All these methods are legit public methods and users should not be bothered if they close the steam twice. This is the choice of this PR.

:+1: 

The GitHub interface doesn't allow me to approve my own PR, even if you wrote most of it. Please approve it and merge it by yourself :)

lpascal-ledger (approved):
Alright!

PR #263, opened 2022-01-06
Update btc app

bigspider:
Added binaries for the latest bitcoin and bitcoin-testnet app (v2.0.1) on firmware 2.1 for Nano S, replacing the previous symlinks.

Had to slightly adjust a test because the strings on screen changed in the new app.


lpascal-ledger (approved):
Sounds good to me

PR #266, opened 2022-01-27
Add env vars to allow different GDB and ELF

Ericson2314:
Our tools are named slightly differently, so this makes it easier to change without needing to fork so hard.


greenknot:
Thanks!

Issue #282, opened 2022-05-02
Werkzeug >= 2.1.0 triggers issues when using the `SpeculosClient`


lpascal-ledger:
I've encountered an unclear bug with a fresh install of Speculos with `pytest`:

```
==================================================================== ERRORS ====================================================================
_______________________________________________________ ERROR at setup of test_app_name ________________________________________________________
tests/speculos/conftest.py:32: in client
    with SpeculosClient(app=str(file_path), args=args) as client:
../../../../.virtualenvs/new/lib/python3.8/site-packages/speculos/client.py:216: in __enter__
    self.start()
../../../../.virtualenvs/new/lib/python3.8/site-packages/speculos/client.py:208: in start
    self.open_stream()
../../../../.virtualenvs/new/lib/python3.8/site-packages/speculos/client.py:73: in open_stream
    check_status_code(self.stream, "/events")
../../../../.virtualenvs/new/lib/python3.8/site-packages/speculos/client.py:30: in check_status_code
    raise ClientException(
E   speculos.client.ClientException: HTTP request on /events failed, status=400, error=b'{"message": "The browser (or proxy) sent a request that this server could not understand."}\n'
------------------------------------------------------------ Captured stdout setup -------------------------------------------------------------
 * Serving Flask app 'speculos.api.api' (lazy loading)
 * Environment: development
 * Debug mode: off
------------------------------------------------------------ Captured stderr setup -------------------------------------------------------------
10:36:41.448:werkzeug:  * Running on all addresses (0.0.0.0)
   WARNING: This is a development server. Do not use it in a production deployment.
 * Running on http://127.0.0.1:5000
 * Running on http://10.0.56.136:5000 (Press CTRL+C to quit)
[*] speculos launcher revision: 
[*] using SDK version 2.1 on nanos
[*] loading CXLIB from "/home/lpascal/.virtualenvs/new/lib/python3.8/site-packages/speculos/cxlib/nanos-cx-2.1.elf"
[*] patching svc instruction at 0x126358
[*] patching svc instruction at 0x400025d8
10:36:41.490:werkzeug: 127.0.0.1 - - [02/May/2022 10:36:41] "GET /events?stream=true HTTP/1.1" 400 -
-------------------------------------------------------------- Captured log setup --------------------------------------------------------------
[...]

PR #285, opened 2022-05-04
[ci][fix] Sharing Python venv between build job and upload job

lpascal-ledger:
Latest merged [PR](https://github.com/LedgerHQ/speculos/pull/284) broke the CI: the python venv is not present in the last step, when uploading the new Speculos package.

This PR upload the venv as an artifact to use it in the last job.

We could use cache, but I'm not sure how the cache mechanism is updated with newer versions (which could easily break in Python, [see this issue](https://github.com/LedgerHQ/speculos/issues/282)) in GitHub, and generally I don't trust caches. I'm open to opinions.

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -119,8 +119,8 @@ jobs:
         name: dist
         path: dist
 
-  deploy_python_package:
-    name: Deploy the Speculos Python package
+  deploy_python_package_and_docker:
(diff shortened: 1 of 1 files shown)


lpascal-ledger:
> This PR upload the venv as an artifact to use it in the last job.-

No: Uploading the venv is too long and bothersome, and having a `venv` as workflow artifact is dirty.
Creating a venv from scratch is easy and fast, so I'm proposing that.

greenknot (approved):

PR #289, opened 2022-05-05
[ci][fix] Speculos docker build and tested before pushed

lpascal-ledger:


lpascal-ledger:
To test the built docker, I propose a dedicated job to build it and start the boilerplate application with it.
The docker is built with multistage rather than relying on previous steps like before - which makes it hard to know what build artifacts are needed or not. The first stage build dependencies, the second one copy them on the final image (apart from the inter-stage copies, this stage is mostly what was existing previously).
Checking that the app is started and can be reached on the APDU port indicates the build is valid.

This job runs everytime the CI is triggered, not only before the docker is pushed (on master), so we always check the docker has not been broken.

@greenknot Do you think there is other resources to add inside the docker, or is there any other check that could be performed?

greenknot (approved):

PR #296, opened 2022-05-17
How to build a Docker speculos container on Mac M1 machine

yogh333:


src/vnc/CMakeLists.txt
```diff
@@ -4,6 +4,10 @@ project(VncServer C)
 
 add_compile_options(-Wall)
 
+if (CMAKE_HOST_SYSTEM_PROCESSOR MATCHES "aarch64")
```
greenknot, line 7: Quick comment: `MATCHES` seems more appropriate for regexp. Maybe use `EQUAL` instead?

yogh333: Yes agreed, replaced by STREQUAL

PR #299, opened 2022-06-09
Enabling of generic curve for starknet

rdubois-crypto:
- Generic curves are handled (previously only supporting known nid curves)
- Test vector added for starkcurve in test_ecdsa
- ECC constants exported to specific files for lisibility purposes


rdubois-crypto:
Modification in accordance to comments performed.

rdubois-crypto:
Push through PR302 with limited changes

PR #301, opened 2022-06-16
disable gdbinit usage for Mac M1

yogh333:
When using gdb with Mac M1, disable gdbinit usage

### DIFF
--- a/tools/debug.sh
+++ b/tools/debug.sh
@@ -24,8 +24,14 @@ function main()
     launcher_text_addr=$(get_text_addr "$launcher_path")
     local gdbinit="$script_dir/gdbinit"
 
-    cat >/tmp/x.gdb<<EOF
+    local arch
+    arch=$(uname -m)
+    if [ "$arch" != "aarch64" ]; then
+        cat >/tmp/x.gdb<<EOF
 source ${gdbinit}
+EOF
+    fi
+    cat >>/tmp/x.gdb<<EOF
 set architecture arm
 target remote 127.0.0.1:1234
 handle SIGILL nostop pass noprint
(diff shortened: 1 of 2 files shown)


tools/debug.sh
```diff
@@ -24,8 +24,13 @@ function main()
     launcher_text_addr=$(get_text_addr "$launcher_path")
     local gdbinit="$script_dir/gdbinit"
 
-    cat >/tmp/x.gdb<<EOF
+    local arch=$(uname -a)
+    if [ "$arch" != "aarch64"]; then
```
greenknot, line 28: Please add a space between `"aarch64"` and `]`. From `shellcheck`:

```
    if [ "$arch" != "aarch64"]; then
    ^-- SC1009: The mentioned parser error was in this if expression.
       ^-- SC1073: Couldn't parse this test expression.
                              ^-- SC1020: You need a space before the ].
                              ^-- SC1072: Missing space before ]. Fix any mentioned problems and try again.
```

tools/debug.sh
```diff
@@ -24,8 +24,13 @@ function main()
     launcher_text_addr=$(get_text_addr "$launcher_path")
     local gdbinit="$script_dir/gdbinit"
 
-    cat >/tmp/x.gdb<<EOF
+    local arch=$(uname -a)
```
greenknot, line 27: Please use the following pattern:

```console
local arch
arch=$(uname -a)
```

From `shellcheck`:

```
    local arch=$(uname -a)
          ^-- SC2155: Declare and assign separately to avoid masking return values.
```

greenknot (changes requested):

tools/debug.sh
yogh333: Done

tools/debug.sh
yogh333: Done

greenknot (approved):
Looks good to me, thanks!

PR #305, opened 2022-06-21
automatically define NATIVE_64BITS on 64-bit CPUs

greenknot:
Predefining NATIVE_64BITS leads to a redefinition if speculos source files are
built with cxlib:

https://github.com/LedgerHQ/nanox-secure-sdk/blob/86324f6a73e269c04f6a3e3cf41f6569a8cc6c6b/lib_cxng/include/lcx_common.h#L38

### DIFF
--- a/src/bolos/cx_utils.h
+++ b/src/bolos/cx_utils.h
@@ -3,6 +3,10 @@
 #include <stdint.h>
 #include <unistd.h>
 
+#if UINTPTR_MAX == UINT64_MAX
+#define NATIVE_64BITS
+#endif
+
 #ifndef NATIVE_64BITS // NO 64BITS
 /** 64bits types, native or by-hands, depending on target and/or compiler
  * support.


src/bolos/cx_utils.h
```diff
@@ -3,6 +3,10 @@
 #include <stdint.h>
 #include <unistd.h>
 
+#if INTPTR_MAX == INT64_MAX
```
niooss-ledger, line 6: I was wondering whether any other project was doing such a check, and actually Musl is doing this :) https://git.musl-libc.org/cgit/musl/tree/include/stdint.h?h=v1.2.3#n105

... but it uses the unsigned type instead: `#if UINTPTR_MAX == UINT64_MAX`. I believe it makes sense to use unsigned types instead (as "pointers" are usually considered unsigned), unless there is a strong argument for using `INTPTR_MAX == INT64_MAX` that I missed.

TL;DR: LGTM, and feel free to change it to `#if UINTPTR_MAX == UINT64_MAX` if you want

niooss-ledger (approved):

src/bolos/cx_utils.h
greenknot: Thanks! (I update macros to unsigned version.)

PR #306, opened 2022-06-22
bump flask-restful minimum version to 0.3.9

aiooss-ledger:
flask-restful 0.3.8 is incompatible with Flask 2.0, or speculos requires at least Flask 2.0.0.


greenknot (approved):
LGTM, thanks @aiooss-ledger!

PR #308, opened 2022-06-23
Add stark256 support

Saltari:
This PR aims at two things:
- Adding the support for the elliptic curve Stark256 (which is already supported by Bolos)
- Adding an API to create generic curves (not having full support (ie: no NID) in OpenSSL)


Saltari:
On the testing side of things:
It was too much of a pain to backport the support of the generic curve API for older SDKs as there is no generic curve domain implementation so no vectors were added to the existing ecdsa unit test.
The implementation was tested using the crypto test app CXT running on speculos while running `ecdsa_sign/verify` operations

src/bolos/cx_ec.c
```diff
@@ -868,6 +868,81 @@ static cx_curve_weierstrass_t const C_cx_BrainpoolP512t1 = {
 
 };
 
+/* ------------------------------------------------------------------------ */
+/* ---                         Stark256                                 --- */
+/* ------------------------------------------------------------------------ */
+
+static uint8_t const C_cx_Stark256_a[] = {
```
greenknot, line 875: The curve parameters are also defined in `cx_ec_domain.c`. Is it possible to define them once and declare them as `extern` in a `.h` file?

src/bolos/cx_ec_domain.c
```diff
@@ -1399,6 +1484,80 @@ int cx_nid_from_curve(cx_curve_t curve)
...
+
+  /* Set up the values for the various parameters */
+  if ((NULL == (a = BN_bin2bn(domain->a, domain->length, NULL))) ||
+      (NULL == (b = BN_bin2bn(domain->b, domain->length, NULL))) ||
+      (NULL == (p = BN_bin2bn(domain->p, domain->length, NULL))) ||
+      (NULL == (order = BN_bin2bn(domain->n, domain->length, NULL))) ||
+      (NULL == (x = BN_bin2bn(domain->Gx, domain->length, NULL))) ||
+      (NULL == (y = BN_bin2bn(domain->Gy, domain->length, NULL)))) {
+    errx(1, "error when setting up the BN CTX parameters");
+    return NULL;
```
greenknot, line 1527: Returning `NULL` here prevent allocated variables (with `BN_bin2bn`) from being freed. This pattern is also present in error handling below.

I'd recommend adding an `error:` label before variables freeing at the end of the function and using `goto error;` instead of `return NULL;`

src/bolos/cx_ec_domain.c
```diff
@@ -23,6 +23,8 @@ static uint8_t const C_cofactor_1[] = { 0, 0, 0, 1 };
 static uint8_t const C_cofactor_4[] = { 0, 0, 0, 4 };
 static uint8_t const C_cofactor_8[] = { 0, 0, 0, 8 };
 
+static int const NID_GENERIC = 0;
```
greenknot, line 26: I think `NID_undef` is already defined to `0` and be used instead of `NID_GENERIC`.

src/bolos/cx_ec_domain.c
```diff
@@ -1399,6 +1484,80 @@ int cx_nid_from_curve(cx_curve_t curve)
   return nid;
 }
 
+EC_GROUP *cx_group_from_nid_and_curve(int nid, cx_curve_t cid)
+{
+  EC_GROUP *group = NULL;
+  /* create a generic curve if generic nid */
+  if (nid != 0) {
```
greenknot, line 1491: Please use `NID_undef` instead of `0`

greenknot (changes requested):
Thanks @Saltari!

I wrote a few minor comments but overall it looks good to me.

src/bolos/cx_ec.c
Saltari: The curve parameters are actually useless in this file as we are not planning on adding the support of Stark on older SDK, removing them :)

src/bolos/cx_ec_domain.c
Saltari: ACL

rgeslain-ledger (approved):

src/bolos/cx_ec_domain.c
```diff
@@ -1399,6 +1482,81 @@ int cx_nid_from_curve(cx_curve_t curve)
...
+{
+  /* Get curve domain to init the EC_GROUP */
+  const cx_curve_domain_t *domain;
+  if (NULL == (domain = cx_ecdomain(cid))) {
+    errx(1, "error when getting the curve domain");
+    goto error;
+  }
+
+  BN_CTX *ctx;
+  EC_GROUP *curve;
```
greenknot, line 1507: I think `curve` should be initialized to `NULL`, I don't know why the compiler doesn't complain.

greenknot (changes requested):
Last requested change

src/bolos/cx_ec_domain.c
Saltari: ACK

greenknot (approved):
Thanks @Saltari! Feel free to merge it.

PR #309, opened 2022-06-23
bolos: introduce cx_common.h and rework header includes

xchapron-ledger:
The cx_common.h header content comes from the nanos-secure-sdk
Its purpose is to describe the flags for the mode parameter of many
bolos crypto functions.
Many of the flags definitions were already present, but they were
scattered in many header and some flags were missing.

In order to fix potential missing include issue with the change of
localisation of the flag definition, the header inclusion have been
reworked with a logic similar to what is used in the nano SDK.


xchapron-ledger:
@greenknot Do you know if there was a reason for speculos cx header specific split? If not, what do you think of these changes?

PR #320, opened 2022-09-01
Update pip files and add pytest

Saltari:
Pipfile was still using `python3.8` when the Dockerfile explicitely uses `python3.9`
Adding `pytest` as a dependency to run tests easily within the container


greenknot (approved):
LGTM, thanks

PR #326, opened 2022-11-21
Add events screen content api.

agrojean-ledger:


speculos/api/events.py
```diff
@@ -27,6 +28,12 @@ def remove_client(self, client):
 
     def broadcast(self, event):
         self.logger.debug(f"events: broadcasting {event} to ({len(self.clients)}) client(s)")
+        if self.screen_content:
+            y_trigger = self.screen_content[-1]["y"] + 10
```
xchapron-ledger, line 32: Can you add a comment to explain the purpose of this `10` and how the value was chosen?

xchapron-ledger (commented):
This looks good to me, and I think this new endpoint will be really useful.
However I don't feel comfortable enough with speculos codebase to be confident to give a real "approval".

speculos/api/events.py
agrojean-ledger: Ok

speculos/api/events.py
```diff
@@ -27,6 +33,14 @@ def remove_client(self, client):
...
     def broadcast(self, event):
         self.logger.debug(f"events: broadcasting {event} to ({len(self.clients)}) client(s)")
+        if self.screen_content:
+            # Reset screen content if new event is not below
+            # the last text line of current screen.
+            y_trigger = self.screen_content[-1]["y"] + MIN_LINES_HEIGHT_DISTANCE
+            if event["y"] <= y_trigger:
+                # Reset screen content
+                self.screen_content = []
+        self.screen_content.append(event)
```
lpascal-ledger, line 43: I don't know what these `event`s are generally speaking, as they're casted as dict which could take whatever form they want, so I'm not comfortable with this code.

If events are always in the form of `{"x": X, "y": Y, ...}`, this code is fine (even if malformed, as I'm not too much into defensive coding I'm cool with a exception poping from here). However this does not seem guaranteed. Is it?

Current attribute naming does not seem to support this: `self.event` is a very generic name, but I don't know. So either event are only `(X, Y)` coordinates and in this case the code should be re-written to highlight this (like `self.events` => `self.screen_events` and `self.screen_content` => `self.current_screen_content`), or the new code may break with legitimate event (`KeyError: 'y'`), and so should manage the cases where `event` is something other than `(X, Y)`.

This would be way clearer with proper typing.

speculos/api/events.py
```diff
@@ -27,6 +33,14 @@ def remove_client(self, client):
 
     def broadcast(self, event):
         self.logger.debug(f"events: broadcasting {event} to ({len(self.clients)}) client(s)")
+        if self.screen_content:
+            # Reset screen content if new event is not below
+            # the last text line of current screen.
+            y_trigger = self.screen_content[-1]["y"] + MIN_LINES_HEIGHT_DISTANCE
+            if event["y"] <= y_trigger:
+                # Reset screen content
```
lpascal-ledger, line 41: This comment is redundant with the previous one.

speculos/api/events.py
```diff
@@ -27,6 +33,14 @@ def remove_client(self, client):
 
     def broadcast(self, event):
         self.logger.debug(f"events: broadcasting {event} to ({len(self.clients)}) client(s)")
+        if self.screen_content:
+            # Reset screen content if new event is not below
+            # the last text line of current screen.
+            y_trigger = self.screen_content[-1]["y"] + MIN_LINES_HEIGHT_DISTANCE
+            if event["y"] <= y_trigger:
```
lpascal-ledger, line 40: This intermediate `y_trigger` variable is superfluous, the code is quite clear and straightforward.

speculos/api/events.py
```diff
@@ -27,6 +33,14 @@ def remove_client(self, client):
 
     def broadcast(self, event):
         self.logger.debug(f"events: broadcasting {event} to ({len(self.clients)}) client(s)")
+        if self.screen_content:
+            # Reset screen content if new event is not below
+            # the last text line of current screen.
```
lpascal-ledger, line 38: Adding in the comment that Y start from the top of the screen would make the following comparison unambiguous.

speculos/client.py
```diff
@@ -82,6 +82,10 @@ def close_stream(self) -> None:
             self.stream.close()
         self.stream = None
 
+    def get_screen_content(self) -> dict:
+        resp = requests.get(f"{self.api_url}/events?screencontent=true")
```
lpascal-ledger, line 86: `screencontent` is ambiguous as it does not reflect it is an exclusive option (it filters the other events). Maybe `current_screen_only`?

speculos/api/events.py
```diff
@@ -7,12 +7,18 @@
 
 from .restful import AppResource
 
+# Approximative minimum vertical distance
```
lpascal-ledger, line 10: Why all the newlines? The CI linter accepts line until 120 characters, this comment goes until 91 char so it's fine.

lpascal-ledger (changes requested):
I'm not sure of the `event` argument. It comes from [this SeproxyHAL attributes](https://github.com/LedgerHQ/speculos/blob/master/speculos/mcu/seproxyhal.py#L148), which is filled from[ here](https://github.com/LedgerHQ/speculos/blob/master/speculos/mcu/seproxyhal.py#L252) or [here](https://github.com/LedgerHQ/speculos/blob/master/speculos/mcu/seproxyhal.py#L263),  and I fear the code of this PR will break in the first case (OCR).

speculos/api/events.py
agrojean-ledger: Updated.

speculos/api/events.py
agrojean-ledger: Done.

speculos/client.py
```diff
@@ -82,6 +82,10 @@ def close_stream(self) -> None:
...
+        return resp.json()
```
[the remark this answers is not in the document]
agrojean-ledger: Added.

speculos/api/events.py
agrojean-ledger: Done.

speculos/api/events.py
agrojean-ledger: I made sure all the events are properly typed as `TextEvent` when generated.

agrojean-ledger:
> I'm not sure of the `event` argument. It comes from [this SeproxyHAL attributes](https://github.com/LedgerHQ/speculos/blob/master/speculos/mcu/seproxyhal.py#L148), which is filled from[ here](https://github.com/LedgerHQ/speculos/blob/master/speculos/mcu/seproxyhal.py#L252) or [here](https://github.com/LedgerHQ/speculos/blob/master/speculos/mcu/seproxyhal.py#L263), and I fear the code of this PR will break in the first case (OCR).

@lpascal-ledger I made sure the events are properly typed as `TextEvent` when generated at different places in the code.

cbouvet-ledger (approved):

greenknot (approved):

greenknot:
Thanks everyone for this code and for the review! :rocket:

lpascal-ledger (approved):

PR #327, opened 2022-11-26
Load entire segment containing `.text`

Ericson2314:
Do this instead of just loading the section itself.

Sections are just for linking purposes and not needed at runtime with a completely linked app. Segments (corresponding to program headers), on the other hand, are the actual important runtime data for what should be loaded. It is therefore more correct to load or not load whole sections than individual segments.

We have successfully tried out various linker tricks, in order to have things like properly initialized globals and relocations to reflect apps install-time address on the flash. This is a small step that makes it easier to use speculos which such experiments.

We would be happy to further improve speculos on this front, e.g. to make it load multiple program headers provided they are adjacent, or also load "hex" files (to be more realistic to how the "production" workflow is without arbitrary ELF tricks). But both of these are more involved, so we figured it was better to start with something very simple.

### DIFF
--- a/speculos/main.py
+++ b/speculos/main.py
@@ -50,10 +50,18 @@ def get_elf_infos(app_path):
     with open(app_path, 'rb') as fp:
         elf = ELFFile(fp)
         text = elf.get_section_by_name('.text')
+        for seg in elf.iter_segments():
+            if seg['p_type'] != 'PT_LOAD':
+                continue
+            if seg.section_in_segment(text):
+                text_seg = seg
+                break;
+        else:
+            raise RuntimeError("No program header with text section!")
         symtab = elf.get_section_by_name('.symtab')
         bss = elf.get_section_by_name('.bss')
-        sh_offset = text['sh_offset']
-        sh_size = text['sh_size']
+        sh_offset = text_seg['p_offset']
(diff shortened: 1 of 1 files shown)


Ericson2314:
CC @yhql because being able to have properly working globals is *especially* nice for Rust, where vtables are super common.

yhql:
@Ericson2314 This looks like a really nice solution :) So this would load `.text`, `.rodata` transparently I guess. Are segments guaranteed to be contiguously stored?
I did some experiments for relocation as well and ended up `data.rel.ro` and `.got` (this one mostly with gcc though), maybe your experiments did too, do you know if these are correctly mapped as well? :)

Ericson2314:
> So this would load `.text`, `.rodata` transparently I guess.

As long as they go in the same segment. We modified the linker script to specify the program headers explicitly rather than let the linker decide so we can be sure that will happen.

> Are segments guaranteed to be contiguously stored?

Unclear, but also this *won't* also get any subsequent *segment*, just the one that contains `.text`.

For these reasons, we would love to follow this up with a more robust solution, but we wanted to start with this one to kick off the conversation / get something done in the short term.

yhql (approved):

Ericson2314:
Thanks!

PR #328, opened 2022-12-01
use image from ghcr.io

nicexe:


niooss-ledger (approved):

lpascal-ledger:
Thanks for the update. The `ledgerhq/speculos` image is quite old indeed.

PR #330, opened 2022-12-09
Propagate app's exit status

yhql:
In order to easily detect whether the emulated app exited successfully or not, it is necessary to use the exit `code` in the actual call to `_exit`.
This is not necessary in most cases as bolos ignores this exit code, but enables easier testing in CI: if the test calls `os_sched_exit` with a non-zero status, the CI can tell it failed directly.

### DIFF
--- a/speculos.py
+++ b/speculos.py
@@ -1,7 +1,8 @@
 #!/usr/bin/env python3
 """Launch Speculos emulator"""
+import sys
 from speculos.main import main
 
 
 if __name__ == "__main__":
-    main()
+    sys.exit(main())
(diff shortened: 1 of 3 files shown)


greenknot (commented):
Quick feedback: would it be possible to make `run_qemu()` [return](https://github.com/LedgerHQ/speculos/blob/46db75849fb5c7ae705cbaf308a7bea9d5b97989/speculos/main.py#L132) the child's pid (the function [prototype](https://github.com/LedgerHQ/speculos/blob/46db75849fb5c7ae705cbaf308a7bea9d5b97989/speculos/main.py#L87) should be modified accordingly in that case)? It'd prevent the introduction of the global variable `QEMU_PID`.

yhql:
Indeed, the previous solution was way too convoluted. Thanks for the advice!

greenknot (approved):

PR #344, opened 2023-02-09
Support for 'screen_clear' syscall

yhql:
Tested on S, X and S+.

### DIFF
--- a/sdk/bolos_syscalls_2.0.h
+++ b/sdk/bolos_syscalls_2.0.h
@@ -152,6 +152,7 @@
 #define SYSCALL_os_deny_protected_flash_ID_IN 0x60009134UL
 #define SYSCALL_os_allow_protected_ram_ID_IN 0x60009269UL
 #define SYSCALL_os_deny_protected_ram_ID_IN 0x600093e9UL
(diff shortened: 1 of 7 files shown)


xchapron-ledger (commented):
I'm not sure about the behavior on Stax as you forward to a sys bagl function. Does it work?
Else we should at least not handle the syscall for Stax.

yhql:
If I read correctly it seems none of the `bagl` syscalls are actually used on Stax due to this line: https://github.com/LedgerHQ/speculos/blob/develop/src/emulate_unified_sdk.c#L27-L29
Unfortunately I did not test because I don't have a Stax-compatible app that does this yet

xchapron-ledger (approved):

xchapron-ledger:
> If I read correctly it seems none of the `bagl` syscalls are actually used on Stax due to this line: https://github.com/LedgerHQ/speculos/blob/develop/src/emulate_unified_sdk.c#L27-L29 Unfortunately I did not test because I don't have a Stax-compatible app that does this yet

My bad, indeed it is still a syscall not handled for Stax. so no change of behavior here.

PR #348, opened 2023-02-16
main.py: Add detection of lib name

xchapron-ledger:


apaillier-ledger (approved):
Tested with Ethereum app and Paraswap plugin. :+1:

PR #372, opened 2023-05-15
Pause timer while OCR is running

fbeutin-ledger:


fbeutin-ledger:
No comments yet, I'll add them later

xchapron-ledger (commented):
Maybe another way of doing it which could be equivalent while avoiding having to introduce a `threading.Condition()` would be simply checking `self.pause` in `add_tick()`.
What do you think about it?

fbeutin-ledger:
> No comments yet, I'll add them later

Done + separated the timer in a proper class

speculos/mcu/seproxyhal.py
```diff
@@ -48,23 +48,61 @@ class SephTag(IntEnum):
...
+        self.stop: bool = False
+        self.pause: bool = False
+        self.resume: threading.Condition = threading.Condition()
+        self.add_tick: Callable = add_tick
 
-    There is no need to prevent clock drift.
-    """
+        """
+        Initializes the Ticker Daemon
+        The goal of this daemon is to emuate the time spent in the application by sending it
```
xchapron-ledger, line 61:
```suggestion
        The goal of this daemon is to emulate the time spent in the application by sending it
```

xchapron-ledger (dismissed):

speculos/mcu/seproxyhal.py
fbeutin-ledger: Done

speculos/mcu/seproxyhal.py
```diff
@@ -48,23 +48,61 @@ class SephTag(IntEnum):
 RENDER_METHOD = RenderMethods(0, 1)
 
 
-def ticker(add_tick):
-    """
-    Function ran in a thread to trigger a tick at a periodic interval.
+class TimeTickerDaemon(threading.Thread):
+    def __init__(self, add_tick, *args, **kwargs):
+        super(TimeTickerDaemon, self).__init__(name="time_ticker", daemon=True)
```
lpascal-ledger, line 53:
```suggestion
        super().__init__(name="time_ticker", daemon=True)
```
No need to explicit the motherclass since Python3.

speculos/mcu/seproxyhal.py
```diff
@@ -48,23 +48,61 @@ class SephTag(IntEnum):
...
+        self.resume: threading.Condition = threading.Condition()
+        self.add_tick: Callable = add_tick
 
-    There is no need to prevent clock drift.
-    """
+        """
+        Initializes the Ticker Daemon
+        The goal of this daemon is to emulate the time spent in the application by sending it
+        ticker events at a regular interval TICKER_DELAY.
+        This daemon can be paused and resumed through it's API to stop the flow of time in the MCU
```
lpascal-ledger, line 62:
```suggestion
        This daemon can be paused and resumed through its API to stop the flow of time in the MCU
```

speculos/mcu/seproxyhal.py
```diff
@@ -48,23 +48,61 @@ class SephTag(IntEnum):
 RENDER_METHOD = RenderMethods(0, 1)
 
 
-def ticker(add_tick):
-    """
-    Function ran in a thread to trigger a tick at a periodic interval.
+class TimeTickerDaemon(threading.Thread):
+    def __init__(self, add_tick, *args, **kwargs):
```
lpascal-ledger, line 52:
```suggestion
    def __init__(self, add_tick: Callable, *args, **kwargs):
```
It's better to type-doc the function signature rather than the object properties as:
- `mypy` will not check the body of the method if the signature is not typed
- also `mypy` could flag unmatching  type when the method is called (`in super function, when instantiating TimeTickerDaemon, the constructor was expecting A but got B`) rather than inside the method itself (`in inner method __init__, property add_tick was expecting A but got B `), which is generally more helpful
- I _think_ most doc tools (such as doxygen) can extract argument types from signature

speculos/mcu/seproxyhal.py
```diff
@@ -48,23 +48,61 @@ class SephTag(IntEnum):
...
-    """
+        """
+        Initializes the Ticker Daemon
+        The goal of this daemon is to emulate the time spent in the application by sending it
+        ticker events at a regular interval TICKER_DELAY.
+        This daemon can be paused and resumed through it's API to stop the flow of time in the MCU
+
+        :param add_tick: The callback function to add a Ticker Event to the packet manager
+        :type add_tick: Backend
+        """
```
lpascal-ledger: Unconventional, to have this doc block after the method. :thinking:

speculos/mcu/seproxyhal.py
```diff
@@ -48,23 +48,61 @@ class SephTag(IntEnum):
...
+
+    def pause_emulated_time(self):
+        """
+        Pause time emulation done by the daemon, no ticker event will be sent until resume
+        """
+        self.pause = True
```
lpascal-ledger, line 72: If `TimeTickerDaemon.pause` is not to be managed directly, explicit it as an internal property, like `TimeTickerDaemon._pause` or `TimeTickerDaemon.__pause`.

speculos/mcu/seproxyhal.py
```diff
@@ -48,23 +48,61 @@ class SephTag(IntEnum):
...
+        Initializes the Ticker Daemon
+        The goal of this daemon is to emulate the time spent in the application by sending it
+        ticker events at a regular interval TICKER_DELAY.
+        This daemon can be paused and resumed through it's API to stop the flow of time in the MCU
+
+        :param add_tick: The callback function to add a Ticker Event to the packet manager
+        :type add_tick: Backend
+        """
+
+    def pause_emulated_time(self):
```
lpascal-ledger, line 68: I think this name is unnecessary long. `TimeTickerDaemon.pause` and `TimeTickerDaemon.resume` seem enough for me. The fact it is emulated is redundant and irrelevant.

speculos/mcu/seproxyhal.py
```diff
@@ -48,23 +48,61 @@ class SephTag(IntEnum):
...
 
+        self.logger.debug("exiting")
 
-class PacketThread(threading.Thread):
-    TICKER_PACKET = (SephTag.TICKER_EVENT, b'')
 
+class PacketDaemon(threading.Thread):
     def __init__(self, s, status_event, *args, **kwargs):
-        super(PacketThread, self).__init__(name="packet", daemon=True)
+        super(PacketDaemon, self).__init__(name="packet", daemon=True)
```
lpascal-ledger, line 101:
```suggestion
        super().__init__(name="packet", daemon=True)
```

speculos/mcu/seproxyhal.py
```diff
@@ -48,23 +48,61 @@ class SephTag(IntEnum):
...
+                self.resume.wait()
 
-    while True:
-        add_tick()
-        time.sleep(TICKER_DELAY)
+    def run(self):
+        """
+        Main thread function
+        """
+        while not self.stop:
```
lpascal-ledger: I don't see where this property is changed.
I see that the `PacketDaemon` class changes its own `.stop` property, but not this `TimeTickerDaemon` class. It's initialized at `False` and that's it?

speculos/mcu/seproxyhal.py
```diff
@@ -48,23 +48,57 @@ class SephTag(IntEnum):
...
-class PacketThread(threading.Thread):
-    TICKER_PACKET = (SephTag.TICKER_EVENT, b'')
 
+class PacketDaemon(threading.Thread):
     def __init__(self, s, status_event, *args, **kwargs):
-        super(PacketThread, self).__init__(name="packet", daemon=True)
+        super(PacketDaemon, self).__init__(name="packet", daemon=True)
         self.s = s
         self.queue_condition = threading.Condition()
         self.queue = []
```
lpascal-ledger, line 104:
```suggestion
        self.queue: List[SephTag, bytes] = []
```

speculos/mcu/seproxyhal.py
fbeutin-ledger: Done

speculos/mcu/seproxyhal.py
fbeutin-ledger: Done

speculos/mcu/seproxyhal.py
fbeutin-ledger: Done

speculos/mcu/seproxyhal.py
fbeutin-ledger: Done

fbeutin-ledger: and done for `resume`

speculos/mcu/seproxyhal.py
fbeutin-ledger: Done

speculos/mcu/seproxyhal.py
fbeutin-ledger: Removed

speculos/mcu/seproxyhal.py
fbeutin-ledger: Done

lpascal-ledger (dismissed):

PR #373, opened 2023-05-15
[Python package] Clearer code?

lpascal-ledger:
## Rationale

This PR does not simplify or restructure (much) the Python package code, but rather aims at highlighting module / class dependencies and showing off interfaces.

Some modules in the Speculos Python code has evolved fast lately, and without formal / explicit structure, it becomes hard to understand the code or the impact of changes.

I've started by typing the code, which is a huge help to explicit what's used where. In the meantime, `mypy` asserts my assumptions and helps revealing unexpected / strange things, circular dependencies and whatnot.

Second step was to extract interfaces from the code in a neutral place (most in the existing `speculos.mcu.display` module, some other in `speculos.mcu.struct` module, and a `speculos.observer` module) and try to change the code so that it relies more on the abstractions rather than concrete classes (that's basically a dependency inversion task, which will help simplifying intricate code).

I think the third step would be to then restructure the code so that the abstractions are stored in sensible place, and concrete classes are lighter and clearer. I'm not sure this step will be finished in this PR, as it may need more scrutiny and may take longer to merge than this one alone (and the bigger the PR, the more painful the rebasing).

Current changes are:

## CI changes

- Independent Python CI with `flake8` (existed already), + `mypy` and `bandit` (this last is neutralized and will not redify the CI, although its output is interesting)
- Changed the reusable workflow so that the Speculos image is built once and not for each device

### Typing

Adding as much typing hints as possible, slightly modifying the code to add clarity

### Architecture

- `ObserverInterface` and `BroadcastInterface` explicit an observer design pattern used through the `AutomationServer` and the `EventsBroadcaster`, and instantiated in the `main` function. The first one is tagged as legacy, so I suggest it should be cleaned for the next stable version. For now, the new interface helps keeping track of the current usage
- Added a `GraphicalLibrary` interface for both `Bagl` and `NBGL` in order to ease their usage. Direct usage of either `bagl` or `nbgl` in the `seproxyhal` module should not occur, they should be hidden inside the `screen` (a `Display` implementation). It is not done currently, it may require more work, but at least `GraphicalLibrary` covers some `FrameBuffer` methods, so that there is no longer code like `screen.nbgl.m.take_screenshot()`, but instead `screen.gl.take_screenshot()`, with `gl` being a generic type (`GraphicalLibrary`, so either `Bagl` or `NBGL`).
  Unfortunately, as some SephTags are NBGL-specifics, the dispatcher in `SeProxyHal.can_read` uses `assert isinstance(screen.gl, NBGL)` which degrades a bit the abstraction.
- _The `Display` interface is used by `TextScreen`, `Headless` and `Screen`, however these implementation are not used consistently as the latter is shadowed by `QtScreen` which uses `Screen` through indirect composition. `mypy` clearly sees the issue, however the code is quite intricate here so not sure I'll flatten the dependencies here._ => this has been answered in the last bullet of this list
- This `Display` interface also implements a sort of unclear observer design pattern too through its `notifiers` property and methods. I've started to define an interface `IODevice` for these notified classes (often defined as `klass`), they seem to be related to instances stored into `ServerArgs` but I've not yet catched the meaning of this.
- Back on the `Display` shenanigans, there is a very unclear way of managing the application graphic interfaces. There currently is 3 options: `Text`, `Headless` and `Qt`. The two firsts are quite similar, the last however, is very different and more complex. These classes are intertwined with the (now called) `IODevice` objects, and `QtObject` for the Qt one, with cross-dependencies all over the place (things like the notifiers and `IODevice.can_read(klass)`). First step I found there was to cut these classes in two: a `DisplayNotifier`, managing the event notifications, and a `Display`, used by the `IODevice`s to interract with the screen. This way, the interfaces are clear: the `main.py` instantiates a `DisplayNotifier` and run its loop, and this `DisplayNotifier` can internally instantiate the correct `Display` and notify the `IODevice`. This is to be improved yet, but still sounds better than previously, where the `QtScreen` was really **not** on the same level as the others (`Headless` and `TextScreen`).


fbeutin-ledger (commented):
First pass only reviewing the workflows

.github/workflows/python_checks.yml
```diff
@@ -0,0 +1,43 @@
...
+
+  mypy:
+    name: Type checking
+    runs-on: ubuntu-latest
+    steps:
+    - name: Clone
+      uses: actions/checkout@v3
+    - run: pip install mypy types-requests types-setuptools PyQt5-stubs
+    - name: Mypy type checking
+      run: mypy speculos || echo 0
```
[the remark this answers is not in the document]
lpascal-ledger: `mypy` still outputs error for now, I didn't want the CI to be red before a first complete clean.

.github/workflows/python_checks.yml
```diff
@@ -0,0 +1,43 @@
...
+
+  bandit:
+    name: Security checking
+    runs-on: ubuntu-latest
+    steps:
+    - name: Clone
+      uses: actions/checkout@v3
+    - run: pip install bandit
+    - name: Bandit security checking
+      run: bandit -r speculos -ll || echo 0
```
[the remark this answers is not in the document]
lpascal-ledger: `bandit` outputs errors. I added it to have security info, but have not managed them yet. I didn't want the CI to be red before we take some time to clean it, it's not fair fror other PRs.

.github/workflows/reusable_ragger_tests_latest_speculos.yml
```diff
@@ -20,19 +20,47 @@ on:
         type: string
 
 jobs:
-  build_application:
+  build_applications:
     name: Build application using the reusable workflow
```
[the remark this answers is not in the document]
lpascal-ledger: Same here, weird... :eyes:

.github/workflows/reusable_ragger_tests_latest_speculos.yml
```diff
@@ -20,19 +20,47 @@ on:
...
+  build_docker_image:
+    name: Build Speculos Docker image
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+    steps:
+    - name: Clone
+      uses: actions/checkout@v3
+      with:
+          ref: ${{ inputs.speculos_app_branch_name }}
```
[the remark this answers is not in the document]
lpascal-ledger: I haven't put much thought in this, the jobs already existed, I just changed the location. But is it necessary to have this info here? Isn't the default behavior to clone-checkout "the current branch"?

.github/workflows/python_checks.yml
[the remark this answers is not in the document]
lpascal-ledger: I've removed this, as all `mypy` errors have been dealt with. :+1:

speculos/mcu/vnc.py
```diff
@@ -60,15 +69,14 @@ def redraw(self, pixels):
...
-    def can_read(self, s, screen):
+    def can_read(self, s: int, screen: DisplayNotifier):
         '''Process a new keyboard or mouse event message from the VNC server.'''
 
-        assert s == self.s.fileno()
-        assert s == self.p.stdout.fileno()
+        assert s == self.fileno
 
         data = b''
         while len(data) != 6:
```
tdejoigny-ledger, line 78: should be interesting to have some comments here to describe all the numeric values used (description of the `data `array).

tdejoigny-ledger (approved):

speculos/mcu/vnc.py
lpascal-ledger: It is an improvement for another PR. I have no idea what could lie inside `data` here, and structural changes do no change the current usage.

speculos/mcu/vnc.py
```diff
@@ -60,15 +69,14 @@ def redraw(self, pixels):
         self.p.stdin.write(buf)
         self.p.stdin.flush()
 
-    def can_read(self, s, screen):
+    def can_read(self, s: int, screen: DisplayNotifier):
```
fbeutin-ledger, line 72: what is `s` ? Replace with a meaningful value

speculos/mcu/vnc.py
```diff
@@ -9,10 +9,17 @@
...
 import subprocess
 import sys
+from typing import IO, Optional, Tuple
 
+from .display import DisplayNotifier, IODevice
 
-class VNC:
-    def __init__(self, port, screen_size, password=None, verbose=False):
+
+class VNC(IODevice):
```
fbeutin-ledger, line 17: Maybe add more comments while it's still fresh, like one line or two for each class to explain what they do etc

speculos/api/api.py
```diff
@@ -1,36 +1,42 @@
...
         self._notify_exit: socket.socket
-        self.s, self._notify_exit = socket.socketpair()
+        self.sock, self._notify_exit = socket.socketpair()
         self.api_port: int = api_port
 
-    def can_read(self, s: int, screen) -> None:
-        assert s == self.s.fileno()
+    @property
+    def file(self):
+        return self.sock
```
fbeutin-ledger, line 36: I feel like `sock` is more explicit ?

speculos/mcu/apdu.py
```diff
@@ -8,39 +10,48 @@
...
+    def can_read(self, s: int, screen: DisplayNotifier):
+        assert self.fileno == s
+        c, _ = self.file.accept()
         self.client = ApduClient(c)
         screen.add_notifier(self.client)
 
-    def forward_to_client(self, packet):
+    def forward_to_client(self, packet: bytes):
         if self.client is not None:
             self.client.forward_to_client(packet)
```
fbeutin-ledger, line 38: I've seen this type of code multiple times, is it possible to move it to IODevice ?

speculos/mcu/apdu.py
```diff
@@ -65,12 +76,12 @@ def recv_packet(self):
     def can_read(self, s, screen):
```
fbeutin-ledger, line 76: Explicit `s` + typing

speculos/mcu/bagl.py
```diff
@@ -155,7 +162,7 @@ def hal_draw_rect(self, color, x, y, width, height):
                 break
 
     @staticmethod
-    def compute_line_width(font_id, width, text, text_encoding):
+    def _compute_line_width(font_id: int, width: int, text: bytes) -> int:
```
fbeutin-ledger, line 165: Not related to this PR, but I'm not really sure why we need to perform line width computation ? It should be done by the device ?

fbeutin-ledger: Ah ok this might be an MCU API actually, that's ugly to redo the code here, if it changes in the device we'll have real differences..

speculos/mcu/apdu.py
fbeutin-ledger: On a second read probably not

.github/workflows/reusable_ragger_tests_latest_speculos.yml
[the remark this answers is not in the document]
lpascal-ledger: The point is that here, the caller repo and the called repo are the same. This workflow is called by another workflow of speculos itself.
My interrogation was if this reusable workflow was useful in the first.

speculos/api/api.py
lpascal-ledger: In this case yes, but it is an concrete implementation of the abstract `IODevice.file` method.
`IODevice` itself could have been renamed as a more precise `SocketDevice`, as **all but one** children use sockets.
Only the `VNC` class does not use a socket, but a `subprocess.PIPE`. That's why I chose to widen the `socket` to a more generic `file`. :shrug:

speculos/api/api.py
```diff
@@ -1,36 +1,42 @@
...
+
+    def can_read(self, s: int, screen: DisplayNotifier) -> None:
```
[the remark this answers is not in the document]
lpascal-ledger: I will rename it `fd` for file descriptor, and further document the `IODevice` mother class.

speculos/mcu/apdu.py
lpascal-ledger: Maybe on the `can_read` method (I feel some socket devices share some equivalent code), but I think this refacto would a bit too complicated for now. It won't be just moving code to some other place, it would need to actually understand it.

speculos/mcu/apdu.py
lpascal-ledger: Hu, maybe you catched a bug here. `screen` should be a `DisplayNotifier`, but `DisplayNotifier` does not have a `forward_to_app` method, that's a `Display` method. Good catch!

speculos/mcu/bagl.py
lpascal-ledger: Honestly, I have no clue!

speculos/mcu/vnc.py
lpascal-ledger: Replaced with `fd`. But I think it can be removed completely. Will propose a dedicated commit with this simplification.

speculos/mcu/vnc.py
lpascal-ledger: I'll document the mother class `IODevice`

fbeutin-ledger (approved):

PR #392, opened 2023-05-31
Add blst library to the docker image

srasoamiaramanana-ledger:


build.Dockerfile
```diff
@@ -33,7 +34,8 @@ RUN \
   wget --quiet https://www.openssl.org/source/openssl-1.1.1k.tar.gz && \
   wget --quiet https://cmocka.org/files/1.1/cmocka-1.1.5.tar.xz && \
   sha256sum --check SHA256SUMS && \
-  rm SHA256SUMS
+  rm SHA256SUMS && \
+  wget --quiet https://github.com/supranational/blst/archive/refs/heads/master.zip
```
xchapron-ledger, line 38: why not the tar.gz to avoid the need for unzip?
```suggestion
  wget --quiet https://github.com/supranational/blst/archive/refs/heads/master.tar.gz
```

niooss-ledger: Also, would it be possible to use a specific version from https://github.com/supranational/blst/releases instead of `master` branch? I believe it would be better not to depend on the `master`/development branch of another project.

srasoamiaramanana-ledger: I tried with the latest tag v0.3.10 which doesn't build with -Werror. Is it acceptable for you to ignore the warnings then ?

niooss-ledger: If you find a commit which fixes the build warning (in the `master` branch), you can use this commit instead of the tag (replacing `master` by the commit hash, in the URL). The main point of my comment is that it is better not to depend on a dynamic reference such as "the current `master` branch", and to instead specify a "known-good" version of this dependency.

xchapron-ledger (approved):

niooss-ledger (approved):

PR #395, opened 2023-06-07
Scarce screenshot publish

fbeutin-ledger:
Only publish screenshot content when the OCR is finished on Stax
Rework the take_screenshot to avoid recreating the image at every REST api refresh

### DIFF
--- a/speculos/api/events.py
+++ b/speculos/api/events.py
@@ -31,13 +31,18 @@ class EventsBroadcaster:
         self.logger.debug("events: client exited")
         self.clients.remove(client)
 
+    def clear_events(self):
+        self.logger.debug("Clearing events")
+        self.screen_content = []
+
     def broadcast(self, event: TextEvent):
-        self.logger.debug(f"events: broadcasting {asdict(event)} to ({len(self.clients)}) client(s)")
         if self.screen_content:
             # Reset screen content if new event is not below the last text line of
             # current screen. Event Y coordinate starts at the top of the screen.
             if event.y <= self.screen_content[-1].y + MIN_LINES_HEIGHT_DISTANCE:
-                self.screen_content = []
+                self.clear_events()
+
+        self.logger.debug("events: broadcasting %s to %d client(s)", asdict(event), len(self.clients))
         self.screen_content.append(event)
         self.events.append(event)
         for client in self.clients:
--- a/speculos/mcu/display.py
+++ b/speculos/mcu/display.py
@@ -82,7 +82,7 @@ class FrameBuffer(ABC):
         self.screenshot = Screenshot(MODELS[model].screen_size)
         # Init published content now, don't wait for the first request
         if self.model == "stax":
-            self.publish_screenshot()
+            self.update_public_screenshot()
 
     def draw_point(self, x, y, color):
         # There are only 2 colors on the Nano S and the Nano X but the one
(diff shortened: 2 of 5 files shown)


fbeutin-ledger:
See https://github.com/LedgerHQ/ragger/pull/132/files for the Ragger implem

speculos/api/events.py
```diff
@@ -31,13 +31,18 @@ def remove_client(self, client):
...
     def broadcast(self, event: TextEvent):
-        self.logger.debug(f"events: broadcasting {asdict(event)} to ({len(self.clients)}) client(s)")
         if self.screen_content:
             # Reset screen content if new event is not below the last text line of
             # current screen. Event Y coordinate starts at the top of the screen.
             if event.y <= self.screen_content[-1].y + MIN_LINES_HEIGHT_DISTANCE:
-                self.screen_content = []
+                self.clear_events()
+
+        self.logger.info(f"events: broadcasting {asdict(event)} to ({len(self.clients)}) client(s)")
```
lpascal-ledger, line 45: Sounds more like `DEBUG` log than `INFO` to me.

+ avoid using f-strings with `logging` package, as it [lazily interpolates string whether it needs to or not](https://docs.python.org/3/howto/logging.html#optimization) (for instance, if it only outputs ERROR, it won't process DEBUG strings). It's cheap optimization, and can be significant with frequently-used functions as `logging` ones.

lpascal-ledger (approved):
Feels like I'll have to rebase #373 :feelthepain:

speculos/api/events.py
fbeutin-ledger: Well it's interesting information that I feel everyone could be interested about when watching the test run
I honestly feel like ragger and speculos are lacking a lot of logs, even logs that are considered debug today
this verbosity is only show if either the test fails or if we explicitly ask `-s`, so I don't feel bad at all about adding logs

fbeutin-ledger: Done

speculos/mcu/display.py
```diff
@@ -81,11 +93,33 @@ def draw_point(self, x, y, color):
...
 
     def screenshot_update_pixels(self):
+        # Someone tampered our self.pixels content.. update the screenshot to reflect it
+        # NANO only
         self.screenshot.update(self.pixels)
 
     def take_screenshot(self):
         return self.screenshot.get_image()
 
+    def publish_screenshot(self):
```
agrojean-ledger, line 103: I would rather call this "save_screenshot" rather than publish as it sounds like it is sent somewhere

speculos/mcu/display.py
```diff
@@ -81,11 +93,33 @@ def draw_point(self, x, y, color):
...
+        if self.model == "stax":
+            screen_size, data = self.take_screenshot()
+            self.published_screenshot_value = _screenshot_to_iobytes_value(screen_size, data)
+        else:
+            # Never called
+            pass
+
+    def take_published_screenshot(self):
```
agrojean-ledger, line 111: I would maybe call this one just `get_screenshot`.

`take` suggests the screenshot is taken when the function is called, which is only true for nano devices

speculos/mcu/seproxyhal.py
```diff
@@ -288,6 +289,24 @@ def can_read(self, s: socket.socket, screen):
...
+                    if self.refreshed:
+                        self.refreshed = False
+
+                        if not screen.nbgl.disable_tesseract:
+                            # Pause flow of time while the OCR is running
+                            self.time_ticker_thread.pause()
+
+                            # Run the OCR
+                            screen.nbgl.m.update_screenshot()
+                            screen_size, image_data = screen.nbgl.m.take_screenshot()
```
agrojean-ledger, line 301: Can't we only call `screen.nbgl.m.take_screenshot()` only once between here and in `screen.nbgl.m.publish_screenshot()` ?

speculos/mcu/display.py
fbeutin-ledger: `get_public_screenshot`

agrojean-ledger (approved):

PR #398, opened 2023-06-13
Fix fonts

dmorais-ledger:
These commits will fix fonts issues with OCR on Nano X & NanoS+


xchapron-ledger (commented):
Can you fix the CI lint issue?

xchapron-ledger:
Should we commit the `.inc` files? Only the `.json` ones are used.

speculos/mcu/ocr.py
```diff
@@ -81,14 +84,51 @@ def display_char(font: bagl_font.Font, char: str) -> None:
...
+def get_json_font(json_name:str) -> Mapping[Char, BitMapChar]:
+    # If no json filename was provided, just return
+    if json_name is None:
+        return None
+
+    # Build the filename by adding api level information
+    json_name += f"-api-level-{OCR.api_level}.json"
+    # Now add the fonts path (JSON files are in speculos/fonts)
+    script_path = os.path.abspath(os.path.dirname(__file__))
+    json_name = os.path.join(script_path, "..", "fonts", json_name)
```
xchapron-ledger, line 96: You need to add speculos/fonts/*.json in MANIFEST.in, else that wqon't work in packaged version of speculos.

xchapron-ledger, line 96: You need to access the ressource following the cx_lib model: https://github.com/LedgerHQ/speculos/blob/master/speculos/main.py#L152

dmorais-ledger:
> Can you fix the CI lint issue?

Sure, I'll do it. :)

dmorais-ledger:
> Should we commit the `.inc` files? Only the `.json` ones are used.

For the same reason I added the .json files to the SDK in API_LEVEL_5 tag, I think it is better to provides the .inc files associated to the .json ones => in case of troubles, people will be able to compare these .inc files with the ones he's currently using in the SDK.

xchapron-ledger:
> 

Why not just comparing the Json file with the one in the SDK then ?
And considering that the Json file is equivalent to the inc file in the SDK ? We should add a CI check for that!

PR #426, opened 2023-10-10
Implement cx_bn_gf2_n_mul()

aido:
`cx_bn_gf2_n_mul()` is currently not implemented on Speculos which causes an error every time `cx_bn_gf2_n_mul()` is called in an app:
````
syscall 0x05000046 not handled
````
 
However, OpenSSL has a function `BN_GF2m_mod_mul()` which can be used to implement `cx_bn_gf2_n_mul()` in Speculos. `BN_GF2m_mod_mul()` can be accessed when OpenSSL is built without the `no-ec2m` option.

### CODE REVIEW

File: sdk/bolos_syscalls_2.0.h [hunk 1]
```diff
@@ -60,6 +60,7 @@
 #define SYSCALL_cx_bn_is_prime_ID_IN 0x6000ef6bUL
 #define SYSCALL_cx_bn_next_prime_ID_IN 0x6000f070UL
 #define SYSCALL_cx_bn_rng_ID_IN 0x6001dd7cUL
+#define SYSCALL_cx_bn_gf2_n_mul_ID_IN 0x05000046UL
```
srasoamiaramanana-ledger, line 63: The syscall is not available on the corresponding device versions (Nano S 2.0.0, Nano S 2.1.0, Nano X 2.0.0, Nano X 2.0.2), so the syscall shouldn't be added to be compliant.
The syscall ID is already in `bolos_syscalls_unified_sdk.h`.

srasoamiaramanana-ledger (changes requested):

src/emulate_2.0.c
```diff
@@ -259,6 +259,9 @@ int emulate_2_0(unsigned long syscall, unsigned long *parameters,
 
     SYSCALL1(cx_bn_next_prime, "(%u)", uint32_t, a);
 
+    SYSCALL5(cx_bn_gf2_n_mul, "(%u, %u, %u, %u, %u)", uint32_t, r, uint32_t, a,
```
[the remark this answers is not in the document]
aido: Fixed in latest push

sdk/bolos_syscalls_2.0.h
aido: Fixed in latest push

srasoamiaramanana-ledger (approved):

PR #440, opened 2023-12-07
Fetch seed / RNG / attestation / user keys / appname / appversion from env at startup then stick with it

lpascal-ledger:
## Feature 1
Previously, seed, RNG and app name & version were fetched from the environment each time they were needed.
When several Speculos instances runs simultaneously in the same shell, the last declared env variable would overwrite the environment variable for every Speculos instance, and thus every Speculos instances would use the same value (same seed, same RNG, same app name / version).

Now these information are fetched once when Speculos starts, and stored internally. Successive Speculos instances no longer overwrite these information for former Speculos instances which are still running.

## Feature 2
Attestation key and user private keys were not configurable.
Now they can be.

## Implementation

These features are implemented in a `environment.c` module which embeds static variables to stores these values.
This module exposes an `init_environment` function, which is called at the start of Speculos (early in `launcher.c:main`, before the emulator is properly started). This function fetches, checks and sets the variables from the environment, or keeps them with their default values if they are not defined.

Module which need these values (`bolos/endorsement.c`, `bolos/cx.c`, `bolos/os.c`, `bolos/os_eip2333.c` and `bolos/os_bip32.c`) can use `env_get_[seed|rng|user_private_key|user_certificate|app_tag]` to access them.


src/seed.c
```diff
@@ -0,0 +1,93 @@
+#include <err.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+
+#include "seed.h"
+
+/* glory promote mansion idle axis finger extra february uncover one trip
```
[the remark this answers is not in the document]
lpascal-ledger: I wondered, it would simplify the code a bit too. I'm not sure if anyone uses Speculos directly, without using the `Python` client.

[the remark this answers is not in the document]
xchapron-ledger: Ha! I Didn't think of that possibility... would it still work considering the change we have done lately?

lpascal-ledger:
@xchapron-ledger  This can be reviewed, I'm going to add unit tests but that's functionally ready.

src/seed.c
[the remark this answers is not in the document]
lpascal-ledger: I'll keep it as it is for now. There are several other things we should cleanup in Speculos which can be breaking change, let's clean them all together for a `1.0.0` version!

src/environment.c
```diff
@@ -0,0 +1,369 @@
...
+/* glory promote mansion idle axis finger extra february uncover one trip
+ * resource lawn turtle enact monster seven myth punch hobby comfort wild raise
+ * skin */
+static uint8_t default_seed[MAX_SEED_SIZE] =
+    "\xb1\x19\x97\xfa\xff\x42\x0a\x33\x1b\xb4\xa4\xff\xdc\x8b\xdc\x8b\xa7\xc0"
+    "\x17\x32\xa9\x9a\x30\xd8\x3d\xbb\xeb\xd4\x69\x66\x6c\x84\xb4\x7d\x09\xd3"
+    "\xf5\xf4\x72\xb3\xb9\x38\x4a\xc6\x34\xbe\xba\x2a\x44\x0b\xa3\x6e\xc7\x66"
+    "\x11\x44\x13\x2f\x35\xe2\x06\x87\x35\x64";
+
+static const char *SEED_ENV_NAME = "SPECULOS_SEED";
```
xchapron-ledger, line 26: I would have expect a `#define` or a `static const char xx[]` but I guess it works.

xchapron-ledger (dismissed):

src/environment.c
```diff
@@ -0,0 +1,369 @@
...
+
+  if (str == NULL) {
+    warnx("using default app name & version");
+    fprintf(stderr, "[*] Default app name: '%s'\n", app_name.name);
+    fprintf(stderr, "[*] Default app version: '%s'\n", app_version.name);
+    return;
+  }
+
+  char *char_ptr = strchr(str, ':');
+  if (char_ptr == NULL) {
```
xchapron-ledger, line 319: There is a missing else line 317 to avoid the double warning?

lpascal-ledger: Hmm you mean `strchr(NULL, ':')` would return `NULL`? The early `return` line 315 should avoid that.

lpascal-ledger:
While doing tests I figured that storing default `user_certificate` was a bit messy, and that it would be clearer if they were always generate from attestation + user private, so there will be small change on sources yet (once I get my crypto right, cause for some reason I struggle getting the same value of old certificates from the default attestation / key...).

src/environment.c
xchapron-ledger: Oh I missed that early return! Sorry

src/environment.c
lpascal-ledger: I could go with it if it's more idiomatic.

xchapron-ledger: https://stackoverflow.com/a/1431607 :man_shrugging:

lpascal-ledger: https://github.com/LedgerHQ/speculos/pull/440/commits/1c47dccaf473278a133ecdc63b7d34a3828cd1c4

src/environment.c
```diff
@@ -0,0 +1,345 @@
...
+
+  p = getenv(SEED_ENV_NAME);
+  if (p != NULL) {
+    size = unhex(actual_seed.seed, sizeof(actual_seed.seed), p, strlen(p));
+    if (size < 0) {
+      warnx("invalid seed passed through %s environment variable",
+            SEED_ENV_NAME);
+      p = NULL;
+    } else {
+      fprintf(stderr, "[*] Seed initialized from environment: '0x%s'\n", p);
```
xchapron-ledger, line 120: Maybe we should not show the seed value here?
In case someone use a real seed and copy paste the logs...

xchapron-ledger (commented):
Can you remove this now noisy log: https://github.com/LedgerHQ/speculos/blob/master/src/bolos/cx_ec.c#L1655

src/environment.c
lpascal-ledger: > In case someone use a real seed and copy paste the logs...
:see_no_evil: 

Right, I guess beside testing purposes it's not very useful.

xchapron-ledger:
The launch output of Speculos get mixed info. I'm wondering if we can do something simple to fix it to ease readability?

lpascal-ledger:
```
$ speculos ~/repos/apps/boilerplate/app/bin/app.elf 
Device model detected from metadata: nanos
17:53:08.094:speculos: Disabling OCR.
Warning: Ignoring XDG_SESSION_TYPE=wayland on Gnome. Use QT_QPA_PLATFORM=wayland to run on Wayland anyway.
[*] speculos launcher revision: 8e9897d
[*] using SDK version 6 on nanos
[*] loading CXLIB from "/home/lpascal/.virtualenvs/speculos/lib/python3.10/site-packages/speculos/cxlib/nanos-cx-2.1.elf"
[*] patching svc instruction at 0x126358
[*] Seed initialized from environment
[*] RNG initialized by default (time(NULL)): '1704905588'
[*] Private key ('ATTESTATION_PRIVATE_KEY') initialized with default value: '0x138fb9b91da745f12977a2b46f0bce2f0418b50fcb76631baf0f08ceefdb5d57'
[*] Private key ('USER_PRIVATE_KEY') initialized with default value: '0xe15e01d47082f0ea4771c99fe312f9d70093c89af47787fdf82e031f6728b710'
[*] User certificate 1 initialized (size: 71)
[*] Private key ('USER_PRIVATE_KEY') initialized with default value: '0xe15e01d47082f0ea4771c99fe312f9d70093c89af47787fdf82e031f6728b710'
 * Serving Flask app 'speculos.api.api'
 * Debug mode: off
17:53:08.302:werkzeug: WARNING: This is a development server. Do not use it in a production deployment. Use a production WSGI server instead.
 * Running on all addresses (0.0.0.0)
 * Running on http://127.0.0.1:5000
 * Running on http://192.168.1.15:5000
17:53:08.302:werkzeug: Press CTRL+C to quit
QPixmap::scaled: Pixmap is a null pixmap
[*] User certificate 2 initialized (size: 71)
[*] Env app name: 'Boilerplate'
[*] Env app version: '2.1.0'
[*] patching svc instruction at 0x4000204a
[*] patching svc instruction at 0x40002058
```

What bothering you? Given the tool and the amount of features, it still seems reasonable to me.

We could drop "default" lines such as `[*] RNG initialized by default (time(NULL)): '1704905588'`, implying it would only print something if not the default behavior?

xchapron-ledger:
> What's bothering you? Given the tool and the amount of features, it still seems reasonable to me.

The fact that there are lines starting with ` * `,  `[*] `, timestamp, .. that are mixed, And would rather have them grouped.
But I'm probably being too picky?

lpascal-ledger:
Right I see. `[*]` comes from the emulator, all the rest is from the `Flask` server and other Python threads. As different programs, I don't know if we could arrange their logs, except by artificially postponing the server startup by one or two seconds.

lpascal-ledger:
Adding all this crypto stuff may have slowed the emulator startup enough for the Flask API to catchup on log on top of it, so this may be more mixed than it was.

xchapron-ledger:
Yeap, not sure that postponing is a good thing. This will slow test everywhere...
But I preferred the previous logs...

xchapron-ledger:
I'd say, let's remove logs that are mostly non necessary, so that output stay relevant, and let's merge.
Seed / RNG / Private Key / Certificate won't be necessary most of the time. Can we make them available only in a specific debug mode?

lpascal-ledger:
```
$ speculos ~/repos/apps/boilerplate/app/bin/app.elf 
Device model detected from metadata: nanos
18:32:27.220:speculos: Disabling OCR.
Warning: Ignoring XDG_SESSION_TYPE=wayland on Gnome. Use QT_QPA_PLATFORM=wayland to run on Wayland anyway.
[*] speculos launcher revision: b253aa4
[*] using SDK version 6 on nanos
[*] loading CXLIB from "/home/lpascal/.virtualenvs/speculos/lib/python3.10/site-packages/speculos/cxlib/nanos-cx-2.1.elf"
[*] patching svc instruction at 0x126358
[*] Seed initialized from environment
 * Serving Flask app 'speculos.api.api'
 * Debug mode: off
18:32:27.368:werkzeug: WARNING: This is a development server. Do not use it in a production deployment. Use a production WSGI server instead.
 * Running on all addresses (0.0.0.0)
 * Running on http://127.0.0.1:5000
 * Running on http://192.168.1.15:5000
18:32:27.368:werkzeug: Press CTRL+C to quit
QPixmap::scaled: Pixmap is a null pixmap
[*] Env app name: 'Boilerplate'
[*] Env app version: '2.1.0'
[*] patching svc instruction at 0x4000204a
[*] patching svc instruction at 0x40002058
```

I've dropped quite some logs. Smaller, yet mixed up.

There's no logging mechanism in the emulator, only `fprintf(stderr, ...)`, so hard to have DEBUG prints except by adding a whole log level mechanism - which is quite heavy to do.

And TBH, most of the logs here are not that interesting: either the values are the default one, and you can easily get them from the code if needed, or the user set them through the environment, and in this case it probably already has them in a variable or something.

xchapron-ledger (approved):

xchapron-ledger (approved):

PR #445, opened 2024-01-18
exchange: add timeout in ticks

sgliner-ledger:


speculos/api/apdu.py
```diff
@@ -55,5 +61,6 @@ def post(self):
         except jsonschema.exceptions.ValidationError as e:
             return {"error": f"{e}"}, 400
 
-        data = bytes.fromhex(args.get("data"))
-        return Response(stream_with_context(self._bridge.exchange(data)), content_type="application/json")
+        tick_timeout = args.get("data")["tick_timeout"]
+        data = bytes.fromhex(args.get("data")["apdu"])
```
xchapron-ledger, line 65: This will break the compatibility for user not using the Speculos python client.
Can you keep the apdu data in the same place and add the itck_timeout somewhere else and with a default value? Not mandatory parameter?

xchapron-ledger (approved):

PR #451, opened 2024-02-07
Cleaning packaging / CI config

lpascal-ledger:
- Moved most of the package configuration from deprecated `setup.py` to `pyproject.toml`
  - except the custom (setuptools-specific) `C` emulator build part, which remains in the `setup.py`  file.
  - added package qualifiers (Python versions, GPL3 license)
  - tool configurations (`mypy`, `flake8`) also moved into `pyproject.toml`.
- Synchronized the dependency package versions into `Pipefile` / Pipfile.lock` from `pyproject.toml` (and `setup.py` before it). Useful as some CI tests run using `pipenv` (and so `Pipefile` and the rest) and dependabot was cringed by some old versions.
- Updated all `actions/checkout@` to `v4`
- Pushed the `sys.exit` from `speculos.py` to `speculos/main.py:main`. The package entrypoint goes directly to `speculos/main.py:main` and I guess it is important that the `speculos` CLI returns useful status codes, instead of `0` always. 
  - Is `speculos.py` ever used ? Given how the entrypoint is configured, `speculos.py` is totally bypassed when using the CLI `speculos`. Is it worth keeping this manual launcher?
- Coverage CI
  - Moved from `fast-checks.yml` to `continuous-intergation-workflow.yml`. This job takes 5min+, it is not fast at all.
  - Fixed the `codecov-action`. Codecov is a fussy action which regularly breaks with unclear error messages. I updated the action version, added a token and explicitly declared the `GITHUB_WORKSPACE` a safe directory for `git`, and it seems to work now. Thank you codecov for taking me a couple of hours again.


xchapron-ledger (dismissed):
A bit sad that we go from one setup.py file to still having a setup.py + having a pyproject.toml :/

Also, shouldn't we bump update the changelog and publish a version? Would allow to make sure that if it breaks something, we know it soon?

lpascal-ledger:
Yeah the custom build `BuildSpeculos` is a bit of a pain to migrate with very few documentation. I tried a few strategy but couldn't make it work.
At least now all tool confs are centralized in the same place and the package is almost entirely declaratively defined.

As for the version, a new one will be deployed on `pypi.org`, only not with a stable release version, and we will be able to test it right now (with `pip install --pre speculos`). I intended to do it on one or two app test suite. You prefer with a stable release?

lpascal-ledger:
I just realized that the `speculos/` Python package source directory coverage is not included into the final `codecov` coverage. I'll take a look at why.

xchapron-ledger:
> As for the version, a new one will be deployed on pypi.org, only not with a stable release version, and we will be able to test it right now (with pip install --pre speculos). I intended to do it on one or two app test suite. You prefer with a stable release?

Probably good enough!

lpascal-ledger:
Alright the coverage is slightly better (not spectacular though), the Python package coverage seems to be included now.

lpascal-ledger:
Tested the new version on functional tests on all devices for `boilerplate`, `MISC` and `passwords`, everything is correctly working (modulo a few `passwords` tests which are known to be unstable locally).

PR #457, opened 2024-02-23
speculos: Fasten graphics handling

xchapron-ledger:
This changes allow to reduce app-boilerplate stax tests run time from 50s to 18s on my setup.
Note that this could be further reduced to 10s if using ragger configuration BACKEND_SCOPE="session".

### DIFF
--- a/speculos/mcu/display.py
+++ b/speculos/mcu/display.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import io
 from abc import ABC, abstractmethod
+from functools import cache
 from PIL import Image
 from socket import socket
 from typing import Any, Dict, IO, List, Optional, Tuple, Union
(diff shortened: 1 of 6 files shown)


cedelavergne-ledger (approved):

speculos/mcu/display.py
```diff
@@ -99,34 +71,67 @@ class FrameBuffer:
...
+        return iobytes.getvalue()
 
     def take_screenshot(self) -> Tuple[Tuple[int, int], bytes]:
-        self.current_screen_size, self.current_data = self.screenshot.get_image()
-        return self.current_screen_size, self.current_data
+        return self.current_screen_size, self._get_image()
 
     def update_screenshot(self) -> None:
-        self.screenshot.update(self.pixels)
+        self.screenshot_pixels = {**self.screenshot_pixels, **self.pixels}
```
lpascal-ledger, line 134:
```suggestion
        self.screenshot_pixels.update(self.pixels)
```

speculos/mcu/vnc.py
```diff
@@ -48,23 +48,25 @@ def file(self) -> IO[bytes]:
         assert self.subprocess.stdout is not None
         return self.subprocess.stdout
 
-    def redraw(self, pixels):
+    def redraw(self, pixels, default_color):
```
lpascal-ledger, line 51: While we're at it!
```suggestion
    def redraw(self, pixels: Dict[Tuple[int, int], int], default_color: int) -> None:
```

lpascal-ledger (approved):
Nice improvement !

speculos/mcu/display.py
xchapron-ledger: Not sure, but in previous version there was a comment saying:
`# Don't call update, replace the object instead`
I trusted it, without testing the performance...

PR #459, opened 2024-02-23
[clean] Better (and faster) shutdown process

lpascal-ledger:


cedelavergne-ledger (dismissed):

speculos/api/api.py
```diff
@@ -23,13 +23,13 @@
 class ApiRunner(IODevice):
     """Run the Speculos API server in a dedicated thread, with a notification when it stops"""
     def __init__(self, api_port: int) -> None:
-        self._app: Flask
+        self._app_wrapper: ApiWrapper
```
xchapron-ledger, line 26: Unused? And mismatch `_api_wrapper` below?

lpascal-ledger: Mismatch true

speculos/mcu/struct.py
```diff
@@ -1,20 +1,15 @@
...
 
 @dataclass
 class Model:
     name: str
-    screen_size: Tuple[int, int]
-    box_position: Tuple[int, int]
-    box_size: Tuple[int, int]
+    screen_size: Pixel
+    box_position: Pixel
+    box_size: Pixel
```
xchapron-ledger, line 12: screen_size and box_size are more about number of pixels (width, height) than a Pixel which is a x and y coordinate. But that works the same.

xchapron-ledger (approved):

xchapron-ledger (approved):

speculos/mcu/struct.py
lpascal-ledger: Indeed. I'm changing it as it may be the same type, but is semantically different to a Pixel.

PR #472, opened 2024-04-08
Ensure run_lib uses the same fonts than main app

cedelavergne-ledger:
Add parse_fonts call in the case of run_lib, and use the same config as main app.


src/launcher.c
```diff
@@ -854,7 +855,8 @@ int main(int argc, char *argv[])
     return 1;
   }
 
-  run_app(MAIN_APP_NAME, NULL, fonts_path != NULL);
+  use_nbgl = fonts_path != NULL;
```
xchapron-ledger, line 858: Would be better line 847 just above 
`  if (fonts_path) {` which could be replaced by `if (use_nbgl)`

cedelavergne-ledger: :+1: Good catch

xchapron-ledger (approved):

PR #483, opened 2024-04-30
[add] Swipe capability on API + SpeculosClient

lpascal-ledger:
- In the API, the finger route can now receive `x2` and `y2` in addition to the required `x` and `y` position.
  When they exists, `x2` and `y2` will be used as the **second** coordinates when the action is "press-and-release". They will be ignored in other cases.
- In the `SpeculosClient`, added a `finger_swipe` method in addition to the `finger_touch`. This new method takes the classic `x` and `y` but also a direction in [`up`, `down`, `right`, `left`] which automatically adds `x2` and `y2` to the API call.

**Pros**
- Higher level `swipe` method, handier than manually filling the `x2`, `y2`

**Cons**
- As is, the API does not propose direct control on `x2` / `y2` as the finger route allows (maybe adding optional `x2`, `y2` on `finger_touch` ? [EDIT] Added this, so that the client has all capabilities)
- `x2` and `y2` overflow / underflow is not checked currently. A swipe left with `x = 0` will do `x2 = -10`.
  I thought it was too much code to add for an error which is easy to detect (the raised error is quite clear) + currently `x` and `y` or not checked either (they just need to be > 0 but 80000 would be valid).

Thoughts?

### DIFF
--- a/speculos/api/finger.py
+++ b/speculos/api/finger.py
@@ -18,11 +18,13 @@ class Finger(SephResource):
         action = args["action"]
         delay = args.get("delay", 0.1)
 
+        x1, y1 = args["x"], args["y"]
         if action == "press-and-release":
-            self.seph.handle_finger(args["x"], args["y"], True)
+            self.seph.handle_finger(x1, y1, True)
             self.seph.handle_wait(delay)
-            self.seph.handle_finger(args["x"], args["y"], False)
+            x2, y2 = args.get("x2", x1), args.get("y2", y1)
+            self.seph.handle_finger(x2, y2, False)
         else:
-            self.seph.handle_finger(args["x"], args["y"], action == "press")
+            self.seph.handle_finger(x1, y1, action == "press")
 
         return {}, 200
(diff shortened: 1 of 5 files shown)


speculos/api/finger.py
```diff
@@ -18,11 +18,13 @@ def post(self):
...
         delay = args.get("delay", 0.1)
 
+        x1, y1 = args["x"], args["y"]
         if action == "press-and-release":
-            self.seph.handle_finger(args["x"], args["y"], True)
+            self.seph.handle_finger(x1, y1, True)
             self.seph.handle_wait(delay)
-            self.seph.handle_finger(args["x"], args["y"], False)
+            x2, y2 = args.get("x2", x1), args.get("y2", y1)
+            self.seph.handle_finger(x2, y2, False)
```
xchapron-ledger, line 26: In @abonnaudet-ledger implementation, there was another `handle_finger` at pos x2, y2 with True, before the release. I guess it's not mandatory?

lpascal-ledger: :thinking: Indeed there's a double `handle_finger` in the `screen.py`, like a complete press-and-release action.
Locally, on the API front it works with only one.

lpascal-ledger: It also seems to work correctly with only one call on `screen.py:ouseReleaseEvent`
 Maybe there's a corner case I don't get @abonnaudet-ledger ?

abonnaudet-ledger: Yes it's not necessary, I can remove it if you want

xchapron-ledger (approved):

PR #494, opened 2024-06-20
Remove the vertical alignment assertion on multiples of 4 on Stax

nroggeman-ledger:
The goal of this PR is to remove the vertical alignment assertion on multiples of 4 on Stax, because this constraint doesn't exist anymore on real Stax.


cedelavergne-ledger (approved):

fbeutin-ledger:
Workflows are broken , fix here https://github.com/LedgerHQ/speculos/pull/495

PR #498, opened 2024-06-27
Fix size of allocated buffer when redrawing vnc framebuffer

nroggeman-ledger:
The goal of this PR is to fix the size of allocated buffer when redrawing vnc framebuffer


meriadec:
Tested :heavy_check_mark: 

weblue

meriadec (approved):

PR #513, opened 2024-09-06
[doc] Added Speculos 'pip' installation rather than only C, source compilation

lpascal-ledger:


README.md
```diff
@@ -18,6 +18,40 @@ Usage example:
 # ... and open a browser on http://127.0.0.1:5000
 ```
 
+## Installation
+
+### From Python `pypi` packages
+
+The easiest, stable way to install Speculos is with `pip`:
```
tdejoigny-ledger, line 25: should we add here that the command needs to be run in speculos repo (root) ?

lpascal-ledger: For source install, sure.

cedelavergne-ledger (approved):

PR #515, opened 2024-09-06
[fix] Moving from Jekyll to Sphinx. RTD theme is safe!

lpascal-ledger:
New theme pushed in #514 was utterly broken.
And I could not easily figure out how the Ruby / Jekyll environment was working (I could not make the theme layouts to be used while serving the site locally).
As pretty much nobody in the team masters Ruby, this PR moves the site generating from Jekyll to Sphinx (yay Python).

I used the markdown plugin to minimize the doc `.md` file modification, and the RTD theme is back on line!


.github/workflows/documentation.yml~
```diff
@@ -0,0 +1,54 @@
+name: Documentation generation & update
```
cedelavergne-ledger, line 1: Are you sure this file should be committed :thinking: ?

docs/index.rst
```diff
@@ -0,0 +1,44 @@
+Speculos documentation
+======================
+
+`Source code <https://github.com/LedgerHQ/speculos>`_
```
cedelavergne-ledger, line 4: Why a trailing `_` ?

lpascal-ledger: That's the RST syntax.
Weird I know

cedelavergne-ledger (approved):

PR #519, opened 2024-09-27
add --nfc flag to use NFC apdu media

yrichard-ledger:


speculos/mcu/seproxyhal.py
```diff
@@ -270,7 +275,10 @@ def __init__(self,
                                                    self.socket_helper.wait_until_tick_is_processed)
         self.time_ticker_thread.start()
 
-        self.usb = usb.USB(self.socket_helper.queue_packet, transport=transport)
+        usb_transport = transport if transport in ['hid', 'u2f'] else 'hid'
```
lpascal-ledger, line 278:
```suggestion
        usb_transport = transport if transport.lower() in ['hid', 'u2f'] else 'hid'
```

lpascal-ledger: Arguably weird to check that here, when `USB.__init__` already has equivalent check. Might as well replace the `raise ValueError` by a `self.transport = Hid(self.send_xfer)` there if fallbacking on HID is the expected behavior.

lpascal-ledger:
Some comments on my latest changes:

- Added the NFC flag in Speculos OS settings on Flex and Stax. We could add an argument to be able to configure it at runtime, but I don't have clear use case for this and didn't though it was worth it for now.
- As `USB` and `NFC` are exclusive (Speculos is started with `NFC` xor `HID` xor `U2F` communication), there is little point instantiating and managing a `.usb` and a `.nfc` variable. As they share similar behavior (with only different names), I merged them under the same `TransportLayer` mother class, and instantiate it from a factory function `build_transport`.
- `HID` and `U2F` classes were managed through composition in `USB`, which led to callback injection, backward dependencies (`HID` and `U2F` inherit from a `Transport` class, but are instrumented through a `USB(TransportLayer)` class (which is no transport) and weird interface (the factory function knows the transport - HID, U2F, NFC- but only deals with NFC and "USB", which itself manages HID and U2F).
   Now `HID` and `U2F` **are** (inheritance) `USBTransport`, which **is** (inheritance) itself a `TransportLayer`, just like `NFC` is also a `TransportLayer`. All can be easily instantiated into the factory function.
- As `--nfc` and `--usb` argument could obstruct each other, and `--usb` feels deprecated now that NFC is available (and maybe bluetooth someday?), I rather added a `--transport` argument superseeding the `--usb` one, which is now flagged as deprecated.

lpascal-ledger (approved):

PR #552, opened 2025-02-27
NVRAM data save and load support

iartemov-ledger:
### Description
It makes a part of "App storage functionality simplification" project ( https://github.com/LedgerHQ/ledger-secure-sdk/pull/865). 
But it seems to me that this change is quite independent, does not break the current stuff and so can be already merged . Unless you have some objections. 

- [x] `--save-nvram`  input parameter is introduced that indicates that the NVRAM write have to be saved
- [x] `--load-nvram`  input parameter is introduced that indicates that a previously saved NVRAM data has to be loaded
- [x] the NVRAM data is saved to and pre-loaded from  `<name>_nvram.bin` (usually `main_nvram.bin`) host file
- [x]  a verification is performed for the write boundaries - not possible outside NVRAM zone defined in linker script 
- [x] tested with `app-boilerplate` locally built in 2 variants
    - [x] using only application storage functionality
    - [x] using also `N_` variables  <= everything is saved and loaded
- [x]  to test more locally:
    - [x] app-boilerplate@master (only `N_`)  - no impact
    - [x] tested locally with `app-exchange` and `pytest test/python/ --capture=tee-sys --tb=short -v --device stax -k "bitcoin and fund_valid_1` - no impact, no NVRAM calls. I do not know if any lib call in any app produces an nvm_write. 
- [x] the use in tests:
    - [x] https://github.com/LedgerHQ/app-boilerplate/blob/357d9eac92503843c6d55f84c5043a7ec53aabcd/tests/conftest.py#L24
    - [x] https://github.com/LedgerHQ/app-security-key/blob/6c19bf04a65bfcf23365739cffa25190c82ac00e/tests/functional/conftest.py#L149

### DIFF
--- a/speculos/main.py
+++ b/speculos/main.py
@@ -14,6 +14,7 @@ import signal
 import socket
 import sys
 import threading
+from dataclasses import dataclass
 from elftools.elf.elffile import ELFFile
 from ledgered.binary import LedgerBinaryApp
 from mnemonic import mnemonic
@@ -34,6 +35,25 @@ from .observer import BroadcastInterface
 from .resources_importer import resources
 
 
+@dataclass
+class ElfInfo:
+    text_offset: int = 0
+    text_size: int = 0
+    stack_addr: int = 0
+    stack_size: int = 0
+    ram_addr: int = 0
+    ram_size: int = 0
+    text_addr: int = 0
+    svc_call_addr: int = 0
+    svc_cx_call_addr: int = 0
+    fonts_addr: int = 0
+    fonts_size: int = 0
+    app_nvram_addr: int = 0
+    app_nvram_size: int = 0
+    cx_ram_addr: int = 0
+    cx_ram_size: int = 0
+
+
 DEFAULT_SEED = ('glory promote mansion idle axis finger extra february uncover one trip resource lawn turtle enact '
                 'monster seven myth punch hobby comfort wild raise skin')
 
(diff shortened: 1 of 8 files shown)


src/launcher.c
```diff
@@ -286,6 +327,7 @@ static void *load_app(char *name)
     warnx("failed to find app \"%s\"", name);
     goto error;
   }
+  printf("app name = %s\n", name);
```
cedelavergne-ledger, line 330: Really needed?

iartemov-ledger: Fixed

cedelavergne-ledger (dismissed):

speculos/main.py
```diff
@@ -98,11 +98,28 @@ def get_elf_infos(app_path, use_bagl):
...
+        else:
+            logger.info("The application does not use NVRAM.")
+
         supp_ram = elf.get_section_by_name('.rfbss')
         ram_addr, ram_size = (supp_ram['sh_addr'], supp_ram['sh_size']) if supp_ram is not None else (0, 0)
     stack_size = estack - stack
     return sh_offset, sh_size, stack, stack_size, ram_addr, ram_size, text_load_addr, \
-        svc_call_addr, svc_cx_call_addr, fonts_addr, fonts_size
+        svc_call_addr, svc_cx_call_addr, fonts_addr, fonts_size, \
+        app_nvram_addr, app_nvram_size
```
lpascal-ledger, line 122: At some point it will be nice to convert this into a `dataclass` or something, instead of returning this huge number of not documented or typed values.

src/launcher.c
lpascal-ledger, line 1: Doesn't it need `#include "launcher.h"`? :thinking:

speculos/main.py
```diff
@@ -98,11 +98,28 @@ def get_elf_infos(app_path, use_bagl):
...
+        envram_symbol = symtab.get_symbol_by_name('_envram')
+        app_nvram_addr = 0
+        app_nvram_size = 0
+        if nvram_data_symbol is not None and (envram_data_symbol is not None or envram_symbol is not None):
+            app_nvram_addr = nvram_data_symbol[0]['st_value']
+            if envram_data_symbol is not None:
+                app_nvram_size = envram_data_symbol[0]['st_value'] - nvram_data_symbol[0]['st_value']
+            else:
+                # Applications built with old SDK version
+                app_nvram_size = envram_symbol[0]['st_value'] - nvram_data_symbol[0]['st_value']
```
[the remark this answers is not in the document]
iartemov-ledger: Applied !

speculos/main.py
iartemov-ledger: Voilà: https://github.com/LedgerHQ/speculos/pull/552/commits/b5227de5b2bafc41f6e3087ae33aa56b308751e3

iartemov-ledger:
@lpascal-ledger proposed to add `--save-nvram` parameter and file path to the both `--save-nvram` and `--load-nvram` in order to ease the NVRAM files handling and comparison. 
The issue with this is the use of apps as libs, as it is done for the `app-exchange` tests where 18 different apps are loaded to Speculos.
It is not clear if there are some use cases when an app launched as library accesses the NVRAM, but we should not deprive us of this possibility. And we cannot provide 18x2 (save/load) file paths ;) 

@iartemov-ledger then proposed to  add `--save-nvram`  and to handle the both, `--save-nvram` and `--load-nvram` as boolean, with `_in` and `_out` file suffixes. 
 `--save-nvram`  has been added, but the file name has not been split as right now it would complicate the tests (see  https://github.com/LedgerHQ/app-boilerplate/blob/357d9eac92503843c6d55f84c5043a7ec53aabcd/tests/conftest.py#L24 and https://github.com/LedgerHQ/app-security-key/blob/6c19bf04a65bfcf23365739cffa25190c82ac00e/tests/functional/conftest.py#L149) . 
And I believe it is still simpler to handle it locally or on CI using files manipulation, if ever needed. 
If it is not the case we will be able to add the suffixes later IMHO.

lpascal-ledger (approved):

lpascal-ledger (approved):

PR #555, opened 2025-03-03
Cleanup logs

cedelavergne-ledger:
- Remove unneeded logs
- Fix `logger.warn` deprecated api
- Add `-S/--silent` option to limit _werkzeug_ logs and remove timestamp from logs
- Add `-f/--finger` to log the Finger press positions

### DIFF
--- a/speculos/api/api.py
+++ b/speculos/api/api.py
@@ -63,7 +63,7 @@ class ApiWrapper:
         static_folder = str(resources.files(__package__) / "static")
         self._app = Flask(__name__, static_url_path="", static_folder=static_folder)
         self._app.env = "development"
-        CORS(self._app, resources={r"*": {"origins": "*"}}, support_credentials=True)
+        CORS(self._app, resources={r"*": {"origins": "*"}}, supports_credentials=True)
 
         screen_kwargs = {"screen": screen}
         seph_kwargs = {"seph": seph}
--- a/speculos/main.py
+++ b/speculos/main.py
@@ -151,7 +151,7 @@ def run_qemu(s1: socket.socket, s2: socket.socket, args: argparse.Namespace, use
         cxlib_args = f'{cxlib}:{sh_offset:#x}:{sh_size:#x}:{sh_load:#x}:{cx_ram_size:#x}:{cx_ram_load:#x}'
         argv += ['-c', cxlib_args]
     else:
-        logger.warn(f"Cx lib {cxlib_filepath} not found")
+        logger.warning(f"Cx lib {cxlib_filepath} not found")
 
     # for NBGL apps, fonts binary file is mandatory
     if not use_bagl:
@@ -314,9 +314,8 @@ def main(prog=None) -> int:
         if binary.sections.target is None:
             logger.error("Device model not detected from elf. Then it must be specified")
             sys.exit(1)
-        else:
-            args.model = "nanosp" if binary.sections.target == "nanos2" else binary.sections.target
-            logger.warn(f"Device model detected from metadata: {args.model}")
+        args.model = "nanosp" if binary.sections.target == "nanos2" else binary.sections.target
+        logger.warning(f"Device model detected from metadata: {args.model}")
 
     # 'bagl' is the default value of the binary.sections.sdk_graphics. We need to
     # manage the cases where it is NOT 'bagl' but the section does not exists yet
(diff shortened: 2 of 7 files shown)


speculos/main.py
```diff
@@ -229,7 +228,14 @@ def run_qemu(s1: socket.socket, s2: socket.socket, args: argparse.Namespace, use
 
 
 def setup_logging(args):
-    logging.basicConfig(level=logging.INFO, format='%(asctime)s.%(msecs)03d:%(name)s: %(message)s', datefmt='%H:%M:%S')
+    if args.silent:
+        logging.basicConfig(level=logging.INFO, format='%(name)s: %(message)s')
+        logger.info("Setting werkzeug WARNING logging level")
+        args.log_level.append("werkzeug:WARNING")
```
lpascal-ledger, line 234: I think werkzeug logs are never interesting and everything under `ERROR` could **always** be ignored (silent flag or not).

speculos/main.py
```diff
@@ -229,7 +228,14 @@ def run_qemu(s1: socket.socket, s2: socket.socket, args: argparse.Namespace, use
 
 
 def setup_logging(args):
-    logging.basicConfig(level=logging.INFO, format='%(asctime)s.%(msecs)03d:%(name)s: %(message)s', datefmt='%H:%M:%S')
+    if args.silent:
+        logging.basicConfig(level=logging.INFO, format='%(name)s: %(message)s')
```
lpascal-ledger, line 232: Why a different log format vs. when not silent? If the previous format is not relevant, we better change it too.

Having a different formatting when `--silent` sounds to me like confusing behavior.

speculos/mcu/seproxyhal.py
```diff
@@ -475,6 +477,8 @@ def handle_finger(self, x: int, y: int, pressed: bool):
         '''Forward finger press/release from the GUI to the app.'''
 
         if pressed:
+            if self.fingers:
+                self.logger.info(f"Finger Press at: {x},{y}")
```
lpascal-ledger, line 481: I know upper/lower case in the existing logs is not very consistent already, but why a capital `P` in `Press`?

speculos/main.py
cedelavergne-ledger: Do you just suggest to set the level to `ERROR` instead of `WARNING`?
Do you rather suggest to put such config with `ERROR` level by default (without condition of any command line argument?

speculos/main.py
```diff
@@ -275,6 +281,9 @@ def main(prog=None) -> int:
                         'either HID (default) or U2F (DEPRECATED, use `--transport` instead)')
     parser.add_argument('-T', '--transport', default=None, choices=('HID', 'U2F', 'NFC'),
                         help='Configure the transport protocol: HID (default), U2F or NFC.')
+    parser.add_argument('-p', '--pki-prod', action='store_true', help='Use production public key for PKI')
+    parser.add_argument('-S', '--silent', action='store_true', help='Limit log traces')
+    parser.add_argument('-f', '--fingers', action='store_true', help='Print finger events')
```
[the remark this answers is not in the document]
cedelavergne-ledger: Ok, I can output X,Y by default

speculos/main.py
cedelavergne-ledger: To keep the timestamp..  Maybe it is interesting for dome other people?
In any cases, I can remove the `msecs`.

speculos/mcu/seproxyhal.py
cedelavergne-ledger: No reasons, I will change that :wink:

speculos/api/api.py
```diff
@@ -3,7 +3,11 @@
 from typing import Any, Dict
 from flask import Flask
 from flask_restful import Api
-from flask_cors import CORS
+try:
+    from flask_cors import CORS
+    CORS_AVAILABLE = True
+except ImportError:
+    CORS_AVAILABLE = False
```
lpascal-ledger, line 10: To confirm if there is an actual issue with the `flask-cors` package
If yes we will need to fix it, but we must not silent this kind of error.
```suggestion
```

lpascal-ledger (changes requested):

speculos/api/api.py
cedelavergne-ledger: Fixed. Caming back to the original code

lpascal-ledger (approved):

PR #556, opened 2025-03-04
[NOISSUE] Set unrestricted CORS for speculos API

mbertin-ledger:
In the context of the DMK sample app (url: https://app.devicesdk.ledger.com/) 
we have added the speculos transport for communicating with it but within this app we had the issue regarding the CORS. 
So I added a new addon to the flask server for authorizing all the CORS.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@
 #
 
 # Building the Speculos environment
-FROM speculos-builder:latest AS builder
+FROM ghcr.io/ledgerhq/speculos-builder:latest AS builder
 
 ADD . /speculos
 WORKDIR /speculos/
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@
 #
 
 # Building the Speculos environment
-FROM ghcr.io/ledgerhq/speculos-builder:latest AS builder
+FROM speculos-builder:latest AS builder
 
 ADD . /speculos
 WORKDIR /speculos/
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -4,7 +4,7 @@
 #
 
 # Building the Speculos environment
-FROM ghcr.io/ledgerhq/speculos-builder:latest AS builder
+FROM speculos-builder:latest AS builder
```
cedelavergne-ledger, line 7: Why did you change that?
It seems it will look for speculos builder locally, instead of pulling it, no?
Not sure of this modification....

speculos/api/api.py
```diff
@@ -87,3 +88,9 @@ def __init__(self,
     def run(self):
         # threaded must be set to allow serving requests along events streaming
         self._app.run(host="0.0.0.0", port=self._port, threaded=True, use_reloader=False)
+
+def after_request_func(response):
+    print("after_request executing!")
+    response.headers.add("Access-Control-Allow-Origin", "*")
+    response.headers.add("X-custom-mathieu", "MATHIEU COUCOU")
```
cedelavergne-ledger, line 95: :stuck_out_tongue_closed_eyes:

Dockerfile
mbertin-ledger: yes I had issue building it. maybe something I didn't understand well ?

speculos/api/api.py
mbertin-ledger: yes...

Dockerfile
mbertin-ledger: or my change when building it locally where not taken into account.

cedelavergne-ledger (approved):

PR #561, opened 2025-03-20
Fixing Rust apps crash and adding Speculos Rust tests

iartemov-ledger:
NVRAM data save and load functionality is not implemented yet for Rust apps.
It is not obvious to do it as the linking/relocation is rather different from the C one. 
So let's do the hotfix at first and avoid doing those operations and also NVRAM write access boundary checking for Rust apps. 

- [x] Rust apps crahs fixed
- [x]  `app-boilerplate-rust` build and test run using Speculos docker added

To follow-up Rust NVRAM save and load support for Rust apps: https://github.com/LedgerHQ/speculos/issues/562

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -186,6 +186,16 @@ jobs:
       test_dir: tests
       speculos_app_branch_name: ${{ github.ref }}
 
+  package_and_test_docker_rust:
+    name: Build and test the Speculos docker with a Rust app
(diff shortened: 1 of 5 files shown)


lpascal-ledger (approved):

lpascal-ledger:
`cargo` is invoked from `Ragger` wtf :sob:

iartemov-ledger:
> `cargo` is invoked from `Ragger` wtf 😭 [EDIT] I remember this PR. What a hard, heavy dependency to have...

I agree. 
And I spent quite some time integrating the Rust/C difference into the Speculos tests :)

lpascal-ledger (approved):

PR #565, opened 2025-03-27
PyQt6 code port

apaillier-ledger:


cedelavergne-ledger (approved):

lpascal-ledger (dismissed):
Could you also update the `README.md`?
I know it is vastly outdated already, but just so that we get cleared of every mention of Qt5.

lpascal-ledger (approved):

PR #586, opened 2025-06-30
Add support for API_LEVEL_24

dmorais-ledger:
Add support for API_LEVEL_24 (IO Revamp)
The corresponding **se-os** branch is **release_05_2025-rc2**


CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.22.1] 2025-06-30
```
yogh333, line 8: Should be `0.23.0` at least. And I wonder if we cannot jump directly to `0.24.0` to be aligned with API_LEVEL_24.
WDYT ?

PR #593, opened 2025-07-16
Use ledgered to determine app parameters from manifest

cedelavergne-ledger:


tdejoigny-ledger (commented):
it seems good. 
A rebase is required following the recent changes for API LEVEL 25 support. 
And then I think that we can release a new version

tdejoigny-ledger (approved):

PR #615, opened 2025-12-01
Add 'libxcb-cursor0' dependency to Debian/Ubuntu installation instructions

bigspider:
On a desktop computer with Ubuntu 25.04, speculos would crash at start until I installed this package.


tdejoigny-ledger:
@bigspider still useful?

PR #660, opened 2026-05-11
remove apex delay

tdejoigny-ledger:
tested [here](https://github.com/LedgerHQ/ledger-app-tester/actions/runs/25050543907)
it seems the tests are still ok without the additional delay

### DIFF
--- a/speculos/mcu/display.py
+++ b/speculos/mcu/display.py
@@ -173,10 +173,13 @@ class FrameBuffer:
         return self._public_screenshot_value
 
     def get_public_screenshot(self) -> bytes:
-        if self.model == "stax" or self.model == "flex":
-            # On Stax/Flex, we only make the screenshot public on the RESTFUL api when it is consistent with events
-            # On top of this, take_screenshot is time consuming on stax/flex, so we'll do as few as possible
-            # We return the value calculated last time update_public_screenshot was called
+        if self.model in ("stax", "flex", "apex_p"):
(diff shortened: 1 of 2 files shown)


speculos/mcu/display.py
```diff
@@ -173,10 +173,13 @@ def public_screenshot_value(self) -> bytes:
         return self._public_screenshot_value
 
     def get_public_screenshot(self) -> bytes:
-        if self.model == "stax" or self.model == "flex":
-            # On Stax/Flex, we only make the screenshot public on the RESTFUL api when it is consistent with events
-            # On top of this, take_screenshot is time consuming on stax/flex, so we'll do as few as possible
-            # We return the value calculated last time update_public_screenshot was called
+        if self.model in ("stax", "flex", "apex_p"):
```
cedelavergne-ledger: Could be rather something like 
`if not self.model.is_nano`
?

tdejoigny-ledger: @cedelavergne-ledger this class (`FrameBuffer`) has not this attribute

cedelavergne-ledger: Using something like
```
from ledgered.devices import Devices
...
if not Devices.get_by_name(self.model).is_nano:
```

This avoid to continue using hardcoded names...

tdejoigny-ledger: ok 👍

cedelavergne-ledger (approved):

